PR #3, opened 2022-04-28
ENG-3606 amend coverage command

jeffgo10:


cecchisandrone:
@jeffgo10 please fix snyk vulnerability

.nycrc
```diff
@@ -0,0 +1,27 @@
+{
+  "cache": false,
+  "check-coverage": true,
+  "extension": [
+    ".ts"
+  ],
+  "include": [
+    "test/**/*.ts"
```
cecchisandrone: coverage should be calculated on `src`, not `test`

package.json
```diff
@@ -33,6 +33,11 @@
   "keywords": [
     "oclif"
   ],
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run coverage"
```
cecchisandrone: this shouldn't be here, move it to `.lintstagedrc`

cecchisandrone (changes requested):

.nycrc
jeffgo10: done

cecchisandrone (approved):

PR #4, opened 2022-04-28
ENG-3559 Implemented init command

zonia3000:


src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+    )
+  }
+
+  private createConfigJson() {
+    // TODO: ENG-3623
+  }
+
+  private createFileFromTemplate(filePath: string, templateFileName: string) {
+    const templateFileContent = fs.readFileSync(
+      path.resolve(__dirname, "..", "resources", templateFileName)
```
zonia3000: I wanted to load the default `Dockerfile` and `.gitignore` from template files located inside sources folder, to avoid hardcoding their content directly inside the class, so I used `__dirname` here.
I've seen a `Do not use "__dirname"` warning and I've read that the preferred method should be using `import.meta.url`. However that construct is not available for commonjs modules (that we specified in the `tsconfig.json`). Can one of our JS experts recommend the best approach for our use case?

package.json
```diff
@@ -17,7 +17,7 @@
     "/oclif.manifest.json"
   ],
   "scripts": {
-    "build": "shx rm -rf dist && tsc -b && npm test",
+    "build": "shx rm -rf dist && tsc -b && shx cp -r src/resources dist/resources && npm test",
```
zonia3000: I added this `cp` command to copy static template files (`Dockerfile` and `.gitignore`) to the `dist` directory. Not sure if it is the best way to handle this.

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "@oclif/core": "1.7.0",
+    "@oclif/errors": "^1.3.5",
```
zonia3000, line 59: I added this in order to raise errors from outside the command class, as described [here](https://oclif.io/docs/commands#thiserrormessage-string--error-options-code-string-exit-number-ref-string-suggestions-string).

src/commands/init.ts
```diff
@@ -0,0 +1,32 @@
...
+    version: Flags.string({ description: "project version" })
+  }
+
+  public async run(): Promise<void> {
+    const { args, flags } = await this.parse(Init)
+
+    const initializer = new Initializer({
+      parentDirectory: process.cwd(),
+      name: args.name as string,
+      version: flags.version ?? "0.0.1"
```
cecchisandrone: extract constant

src/commands/init.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+
+    // Displaying spinner while performing the scaffolding
+    CliUx.ux.action.start(`Initializing new bundle project ${args.name}`)
```
cecchisandrone: Initializing a new bundle project

src/models/bundle-descriptor.ts
```diff
@@ -0,0 +1,32 @@
+export interface Microservice {
```
cecchisandrone: MicroService

src/models/bundle-descriptor.ts
```diff
@@ -0,0 +1,32 @@
...
+  /** Docker image name. This field is computed at packaging phase */
+  image: string
+  dbms: string
+  ingressPath: string
+  healthCheckPath: string
+  roles: string[]
+  env: any
+}
+
+export interface Microfrontend {
```
cecchisandrone: MicroFrontend

src/commands/init.ts
```diff
@@ -0,0 +1,32 @@
+import { CliUx, Command, Flags } from "@oclif/core"
+import Initializer from "../services/initializer"
+
+export default class Init extends Command {
+  static description = "performs the scaffolding of an empty bundle project"
```
cecchisandrone: Performs the scaffolding of a bundle project (we'll add the possibility to init from hub later)

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+import { BundleDescriptor } from "../models/bundle-descriptor"
+
+export interface InitializerOptions {
+  parentDirectory: string
+  name: string
+  version: string
+}
+
+/** Handles the scaffolding of a project bundle */
+export default class Initializer {
```
cecchisandrone: BundleProjectInitializer

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+  private readonly options: InitializerOptions
+
+  constructor(options: InitializerOptions) {
+    this.options = options
+  }
+
+  public async performScaffolding(): Promise<void> {
+    if (!/^[\w-]+$/.test(this.options.name)) {
```
cecchisandrone: extract constant

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+      throw new CLIError(`"${this.options.name}" is not a valid bundle name`)
```
cecchisandrone: Give a hint about supported characters

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.createGitignore()
+    this.createConfigJson()
+    this.initGitRepo()
+  }
+
+  private createBundleDirectories() {
+    const bundleDir = this.getBundleDirectory()
+
+    if (fs.existsSync(bundleDir)) {
+      throw new CLIError(`${bundleDir} already exists`)
```
cecchisandrone: `Directory ${bundleDir} already exists`

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    try {
```
cecchisandrone, line 41: Shouldn't this be before the previous check?

src/services/initializer.ts
cecchisandrone: @Dv92 this is for you

package.json
cecchisandrone: I think we may need a separated script to copy resources. @Dv92 any hint on this?

test/commands/init.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    .stub(
+      fs,
+      "accessSync",
+      sinon.stub().throws({ stderr: "permission denied" })
+    )
+    .command(["init", "bundle-no-permission"])
+    .catch(error => {
+      expect(error.message).to.contain("is not writable")
+    })
+    .it("handles parent directory not writable")
```
cecchisandrone: handles a not writable parent directory

cecchisandrone (changes requested):

src/services/initializer.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    fs.mkdirSync(bundleDir)
+    fs.mkdirSync(this.getBundleFilePath(".ent"))
+    fs.mkdirSync(this.getBundleFilePath(".ent", "output"))
+    fs.mkdirSync(this.getBundleFilePath("microservices"))
+    fs.mkdirSync(this.getBundleFilePath("microfrontends"))
+    fs.mkdirSync(this.getBundleFilePath("epc"))
+  }
+
+  private createBundleDescriptor() {
```
cecchisandrone, line 61: It's worth to move this logic to a dedicated service in order to implement the CRUD there. Wdyt?

src/services/initializer.ts
Dv92: I think for our current context (Node.js 14) we can ignore this warning for this reason:

`import.meta` is introduced by ESM for retrieving information about the current module (is the future way) and in the Node.js context will be a replacement for `__dirname` and `__filename`, there is some problem for our case:

Node.js 14 ESM supports is experimental (https://nodejs.org/dist./v14.10.0/docs/api/esm.html#esm_modules_ecmascript_modules) and not predictable.
`import.meta `is interpreted at runtime, in our case the ESM is purely static because we are going to transpile import to common.js modules so when the program run we effectively don't use ESM.

My advice here is to continue to use `__dirname` and `__filename`, when we schedule future updates of node.js for Entando we can think about changing the runtime module system to be compliant with the future.

cecchisandrone: How can we ignore this in the linter (globally)?

package.json
Dv92: yes, I agree you can make an isolated command

```json
{
  "scripts": {
    "copy:resources": "shx cp -r src/resources dist/resources",
    "build": "shx rm -rf dist && tsc -b && npm run copy:resources && npm test"
  }
}
```

maybe it's better to move outside of `src` the resources dir if this dir contains only static files

src/services/initializer.ts
Dv92: add to `.eslintrc` this rule:

```json
"rules": {
   "unicorn/prefer-module": "off"
}
```

src/services/initializer.ts
zonia3000: Yes, it could be useful for update operations done by other classes. I'm adding it

src/services/initializer.ts
zonia3000: Thanks!

cecchisandrone (approved):

PR #5, opened 2022-05-02
ENG-3623 Add config.json file service

alepintus:


src/services/config/config-service.ts
```diff
@@ -0,0 +1,77 @@
...
+      encoding: "utf-8"
+    })
+    console.log("Default Config successfully read from disk")
+    return JSON.parse(fileContents)
+  }
+
+  writeConfigFile = (bundlePath: string, fileFolder = CONFIG_FOLDER) : void => {
+    const filePath = bundlePath.endsWith("/") ?
+      `${bundlePath}${fileFolder}` :
+      `${bundlePath}/${fileFolder}`
```
Dv92: if you want you can doit also without `if` and string interpolation, like this:

```javascript
const filePath = [bundlePath.split("/").filter(Boolean), fileFolder].flat().join("/")
```

Dv92 (approved):

src/services/config/config-service.ts
alepintus: Hi @Dv92 , 
thanks for the review, I changed it with:
 ` const filePath = path.join(bundlePath, fileFolder)`
 in order to avoid issues with different operating systems.

Dv92 (approved):

src/services/config/config-service.ts
```diff
@@ -0,0 +1,68 @@
+import * as fs from "node:fs"
+import * as path from "node:path"
+import { CLIError } from "@oclif/core/lib/errors"
+
+const DEFAULT_CONFIG_PATH = path.resolve(__dirname)
+const DEFAULT_CONFIG_FILE = "default-config.json"
+const CONFIG_FOLDER = ".ent"
```
cecchisandrone: This is also used in `initializer.ts`. I propose to centralize this constant in a `Paths.ts` class as it will be used in other contexts

src/services/config/default-config.json
```diff
@@ -0,0 +1,4 @@
+{
```
cecchisandrone: This should be moved in `resources` folder

src/services/config/config-service.ts
```diff
@@ -0,0 +1,68 @@
...
+const CONFIG_FILE = "config.json"
+
+export default class ConfigService {
+  private readonly _config: any
+
+  constructor(filePath = DEFAULT_CONFIG_PATH) {
+    this._config = this.readConfigFile(filePath)
+  }
+
+  get config(): any {
```
cecchisandrone: this is not needed

src/services/config/config-service.ts
```diff
@@ -0,0 +1,68 @@
...
+      path.join(filePath, DEFAULT_CONFIG_FILE),
+      {
+        encoding: "utf-8"
+      }
+    )
+
+    return JSON.parse(templateFileContent)
+  }
+
+  writeConfigFile = (bundlePath: string, fileFolder = CONFIG_FOLDER): void => {
```
cecchisandrone: there should be the following methods:
initConfigFile(path)
readConfigFile()
writeConfigFile()

src/services/config/config-service.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    fs.writeFileSync(
+      path.join(filePath, CONFIG_FILE),
+      JSON.stringify(this._config, null, 2),
+      "utf8"
+    )
+  }
+
+  addProperty = (key: string, value: string): void => {
```
cecchisandrone: missing getProperty

src/services/config-service.ts
```diff
@@ -0,0 +1,73 @@
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+import * as os from 'node:os'
+import { CLIError } from '@oclif/core/lib/errors'
+import { CONFIG_FOLDER, CONFIG_FILE } from '../paths'
+
+export default class ConfigService {
+  private _config: any
```
zonia3000: If we know that the format is always a key/value map of strings we may replace that any with something more specific:

    private _config: {[key:string]: string} = {}

alepintus: @zonia3000 thanks for the review. I'm not sure about that, @cecchisandrone WDYT?

cecchisandrone: yes, for the moment it's a reasonable choice

src/paths.ts
```diff
@@ -0,0 +1,5 @@
+export const RESOURCES_FOLDER = 'resources'
+export const CONFIG_FOLDER = '.ent'
+export const TMP_TEST_FOLDER = 'entando-bundle-cli-test'
```
cecchisandrone: better moving test paths in a test class

src/services/config-service.ts
alepintus: Done

cecchisandrone (commented):
Can you also change this:
image
in "Bundle project name"

cecchisandrone (approved):

PR #6, opened 2022-05-04
ENG-3552 Implement mfe add command

raxkaynan:


src/services/microfrontends-manager.ts
```diff
@@ -0,0 +1,47 @@
+import * as path from 'node:path'
+import * as fs from 'node:fs'
+import { CLIError } from '@oclif/errors'
+import { BundleDescriptor, MicroFrontend } from '../models/bundle-descriptor'
+import BundleDescriptorManager from './bundle-descriptor-manager'
+
+const MICROFRONTENDS_DIRNAME = 'microfrontends'
+
+export default class MicroFrontendsManager {
```
Dv92: I suggest you, in general, avoid export default in favor of named export, on the scale can be problematic for [several reasons](https://basarat.gitbook.io/typescript/main-1/defaultisbad). If you use named export can take the advantage of better navigation from the IDE, Typo protection, and class name protection. @cecchisandrone

package.json
```diff
@@ -49,8 +49,8 @@
     ],
     "topicSeparator": " ",
     "topics": {
-      "hello": {
-        "description": "Say hello to the world and others"
+      "mfe": {
+        "description": "Performs microfrontend operations in the current bundle"
```
cecchisandrone: Just "Micro Frontend operations"

src/commands/mfe/add.ts
```diff
@@ -0,0 +1,51 @@
...
+import BundleVerifier from '../../services/bundle-verifier'
+import MicroFrontendsManager from '../../services/microfrontends-manager'
+
+enum Stack {
+  React = 'react',
+  Angular = 'angular'
+}
+
+export default class Add extends Command {
+  static description = 'Adds a microfrontend component to the current bundle'
```
cecchisandrone: "Adds a Micro Frontend component to the bundle"

src/commands/mfe/add.ts
```diff
@@ -0,0 +1,51 @@
...
+      description: 'frontend stack',
+      options: [Stack.React, Stack.Angular],
+      default: Stack.React
+    })
+  }
+
+  static args = [
+    {
+      name: 'name',
+      description: 'name of the microfrontend component',
```
cecchisandrone: "Name of the Micro Frontend component"

src/services/bundle-verifier.ts
```diff
@@ -0,0 +1,20 @@
+import { CLIError } from '@oclif/errors'
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+
+export default class BundleVerifier {
```
cecchisandrone: 1. don't use default export as @Dv92 suggested
2. I would call it BundleService, we may add other functions later

test/commands/mfe/add.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  beforeEach(() => {
+    bundleDescriptorManager.writeBundleDescriptor(bundleDescriptor)
+  })
+
+  after(() => {
+    fs.rmSync(path.resolve(tmpDir), { recursive: true, force: true })
+  })
+
+  test
+    .command(['mfe add', 'default-stack-mfe'])
```
cecchisandrone, line 43: Also provide some tests covering an already existing mfe1, adding mfe2

src/services/microfrontends-manager.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (fs.existsSync(newMfeDir)) {
+      throw new CLIError(`Directory ${newMfeDir} already exists`)
+    }
+
+    fs.mkdirSync(newMfeDir)
+  }
+
+  private addMicroFrontendDescriptor(mfe: MicroFrontend): void {
+    const bundleDescriptor: BundleDescriptor =
```
cecchisandrone: Does this also manage case in which already existing microfrontend with same name exists (we may have that condition if folder is manually removed for example)

src/services/microfrontends-service.ts
```diff
@@ -0,0 +1,55 @@
+import * as path from 'node:path'
+import * as fs from 'node:fs'
+import { CLIError } from '@oclif/errors'
+import { BundleDescriptor, MicroFrontend } from '../models/bundle-descriptor'
+import BundleDescriptorService from './bundle-descriptor-service'
+
+const MICROFRONTENDS_DIRNAME = 'microfrontends'
+
+export class MicroFrontendsService {
```
cecchisandrone: better `MicroFrontendService` (also rename file)

src/services/microfrontends-service.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  private addMicroFrontendDescriptor(mfe: MicroFrontend): void {
+    const bundleDescriptor: BundleDescriptor =
+      this.bundleDescriptorService.getBundleDescriptor()
+    const { microfrontends } = bundleDescriptor
+
+    if (microfrontends.some(({ name }) => name === mfe.name)) {
+      throw new CLIError(
+        `${mfe.name} already exists in the microfrontends section of the bundle descriptor`
```
cecchisandrone: Bundle should be Capitalized as it's an Entando entity (please check in other output strings)

src/commands/mfe/add.ts
```diff
@@ -0,0 +1,51 @@
...
+    const { args, flags } = await this.parse(Add)
+
+    const microFrontend: MicroFrontend = <MicroFrontend>{
+      name: args.name,
+      stack: flags.stack
+    }
+    const microFrontendsService: MicroFrontendsService =
+      new MicroFrontendsService()
+
+    CliUx.ux.action.start(`Adding a new microfrontend ${args.name}`)
```
cecchisandrone: Micro Frontend

src/commands/mfe/add.ts
```diff
@@ -0,0 +1,51 @@
...
+  static description = 'Adds a Micro Frontend component to the bundle'
+
+  static examples = [
+    '$ entando-bundle-cli mfe add my-mfe',
+    '$ entando-bundle-cli mfe add my-mfe --stack react'
+  ]
+
+  static flags = {
+    stack: Flags.string({
+      description: 'frontend stack',
```
cecchisandrone: Micro Frontend stack

src/services/microfrontends-service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  constructor() {
+    this.microfrontendsPath = path.resolve(
+      process.cwd(),
+      MICROFRONTENDS_DIRNAME
+    )
+    this.bundleDescriptorService = new BundleDescriptorService(process.cwd())
+  }
+
+  public addMicroFrontend(mfe: MicroFrontend): void {
```
cecchisandrone: Let's perform some sanitization of the mfe name (use the same regex used in initializer-service) in order to avoid bad chars in the mfe name

cecchisandrone (approved):

PR #8, opened 2022-05-05
ENG-3643 Added debug logging utility

zonia3000:
Oclif [suggests](https://oclif.io/docs/debugging) to use the `debug` module for debug logging, since it is the one that it uses internally. I didn't add the dependency, since we can use the inherited one, however I had to add a rule to the linter to make it working.

I created a debugFactory function that can be used to define loggers inside the classes:

    private static debug = debugFactory(MyClass)

I named it `debug`, as done by the library, to avoid confusion with the `log()` method provided by Oclif `Command` class.

By default the library uses a colorful output in TTY mode and prepends timestamps in non TTY mode.

I didn't find a way to enable a global flag. I created a [custom command base class](https://oclif.io/docs/base_class) defining that flag. So all new commands have to inherit from this one.

Debug library can be activated using the environment variable `DEBUG`, which can take filters (`*` or specific logger namespaces). So `./bin/dev init mybundle -d` is equivalent to `DEBUG=entando-bundle-cli:\* ./bin/dev init mybundle`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ _See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v5.1.1
 
 ## `entando-bundle-cli init NAME`
 
-Performs the scaffolding of a bundle project (we'll add the possibility to init from hub later)
+Performs the scaffolding of a Bundle project (we'll add the possibility to init from hub later)
 
 ```
 USAGE
@@ -73,7 +73,7 @@ FLAGS
   --version=<value>  project version
 
 DESCRIPTION
-  Performs the scaffolding of a bundle project (we'll add the possibility to init from hub later)
+  Performs the scaffolding of a Bundle project (we'll add the possibility to init from hub later)
 
 EXAMPLES
   $ entando-bundle-cli init my-bundle
@@ -96,6 +96,8 @@ nvm use
 
 If you are receiving `command not found` errors while executing pre-commit hooks with Husky using `nvm`, please refer to https://typicode.github.io/husky/#/?id=command-not-found
 
+## Debugging
+
 To print debug information define a static `debug` function using the `debugFactory`:
 
 ```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -60,7 +60,7 @@ _See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v5.1.1
 
 ## `entando-bundle-cli init NAME`
 
-performs the scaffolding of an empty bundle project
+Performs the scaffolding of a bundle project (we'll add the possibility to init from hub later)
```
cecchisandrone: let's always use Bundle in messages as it's a proprietary Entando entity

README.md
```diff
@@ -95,3 +95,15 @@ nvm use
 ```
 
 If you are receiving `command not found` errors while executing pre-commit hooks with Husky using `nvm`, please refer to https://typicode.github.io/husky/#/?id=command-not-found
+
+To print debug information define a static `debug` function using the `debugFactory`:
```
cecchisandrone, line 101: Define a subsection for this and also provide an example to redirect logs to file

src/services/debug-factory-service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default function debugFactory(
+  caller: { name: string } | string
+): (message: string, ...args: any[]) => void {
+  const namespace = typeof caller === "string" ? caller : caller.name
+  const extendedDebugger = rootDebugger.extend(namespace)
+
+  return (message: string, ...args: any[]) => {
+    extendedDebugger.enabled =
+      extendedDebugger.enabled || process.env.ENTANDO_CLI_DEBUG === "true"
```
cecchisandrone: better using ENTANDO_BUNDLE_CLI_DEBUG

cecchisandrone (approved):

PR #10, opened 2022-05-12
ENG-3555 Implement list command

raxkaynan:


src/commands/list.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  static examples = [
+    '$ entando-bundle-cli list',
+    '$ entando-bundle-cli list --ms',
+    '$ entando-bundle-cli list --ms --mfe'
+  ]
+
+  static flags = {
+    ms: Flags.boolean({
+      description: 'List all Micro Service components'
```
cecchisandrone: List only Micro Service components

src/commands/list.ts
```diff
@@ -0,0 +1,54 @@
...
+        components.push(...componentDescriptorService.getMsComponents())
+      }
+
+      if (flags.mfe) {
+        components.push(...componentDescriptorService.getMfeComponents())
+      }
+    }
+
+    const compKeys: Array<string & keyof Component> = [
+      'name',
```
cecchisandrone, line 48: Use uppercase for table headers

src/services/component-descriptor-service.ts
```diff
@@ -0,0 +1,45 @@
+import { BundleDescriptor } from '../models/bundle-descriptor'
+import { Component, ComponentType } from '../models/component'
+import BundleDescriptorService from './bundle-descriptor-service'
+
+export class ComponentDescriptorService {
```
cecchisandrone, line 3: As the name indicates, this class is responsible only for a single component (e.g. getting its version). You put here other operations that should not be responsibility of this class. I would create instead another service class named `ComponentService`, that will be using `BundleDescriptorService` and `ComponentDescriptorService`

src/services/component-descriptor-service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  constructor() {
+    this.bundleDescriptorService = new BundleDescriptorService(process.cwd())
+  }
+
+  getComponents(): Array<Component> {
+    return [...this.getMsComponents(), ...this.getMfeComponents()]
+  }
+
+  getMsComponents(): Array<Component> {
```
cecchisandrone: it's enough to have one parametrized getComponents(type : ComponentType)

cecchisandrone (approved):

PR #12, opened 2022-05-17
ENG-3712 Implement getComponentVersion

raxkaynan:
Implements the method by extracting the actual version of a component; Refactors ComponentService


src/models/component.ts
```diff
@@ -1,11 +1,21 @@
...
   MICROSERVICE = 'microservice'
 }
 
-export interface Component extends Record<string, unknown> {
+export interface PartialComponent {
   name: string
+  stack: string
+}
+
+export interface Component extends PartialComponent, Record<string, unknown> {
```
cecchisandrone: I suggest to keep `Component` as a clean type mapping what we have in the descriptor. If you need smth custom for visualization (table) better extending from it and creating a `VersionedComponent`

src/services/component-descriptor-service.ts
```diff
@@ -1,9 +1,43 @@
...
+import { CLIError } from '@oclif/errors'
+import * as xml2js from 'xml2js'
+import { ComponentType, PartialComponent, Stack } from '../models/component'
 
 export class ComponentDescriptorService {
-  // TODO: implement proper way of getting a component's version
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  getComponentVersion(name: string, type: ComponentType): string {
-    return '0.0.1'
+  public getComponentVersion(
```
cecchisandrone: Make this function take only `Component` as parameter

src/services/component-descriptor-service.ts
```diff
@@ -1,9 +1,43 @@
...
+    const { name, stack } = component
+    const compPath: string = path.resolve(process.cwd(), `${type}s`, name)
+
+    const version: string | undefined =
+      stack === Stack.Node || stack === Stack.React || stack === Stack.Angular ?
+        this.parsePackageJSON(compPath).version :
+        this.parsePomXML(compPath).project?.version?.[0]
+
+    if (!version) {
+      throw new CLIError(`Failed to get version of ${name} ${type}`)
```
cecchisandrone: I would not throw error in case of missing version. Better returning `unkown` for the moment

src/services/component-descriptor-service.ts
```diff
@@ -1,9 +1,43 @@
...
+    }
+
+    return version
+  }
+
+  private parsePackageJSON(dir: string): { version: string } {
+    return JSON.parse(
+      fs.readFileSync(path.resolve(dir, 'package.json'), 'utf-8')
```
cecchisandrone: Handle exception if file is not found/readable. In such case we fallback to `unknown` for the moment

src/services/component-descriptor-service.ts
```diff
@@ -1,9 +1,43 @@
...
+    )
+  }
+
+  private parsePomXML(dir: string): Partial<{ project: { version: string } }> {
+    let result: Partial<{ project: { version: string } }> = {}
+    xml2js.parseString(
+      fs.readFileSync(path.resolve(dir, 'pom.xml'), 'utf-8'),
```
cecchisandrone: Same as above

cecchisandrone (approved):

PR #13, opened 2022-05-17
ENG-3597 JSON to YAML bundle descriptor conversion

zonia3000:


src/paths.ts
```diff
@@ -1,4 +1,8 @@
+import * as path from 'node:path'
+
 export const RESOURCES_FOLDER = 'resources'
 export const CONFIG_FOLDER = '.ent'
 export const CONFIG_FILE = 'config.json'
 export const DEFAULT_CONFIG_FILE = 'default-config.json'
+export const OUTPUT_FOLDER = path.join(CONFIG_FOLDER, 'output')
+export const DESCRIPTORS_FOLDER = path.join(OUTPUT_FOLDER, 'descriptors')
```
cecchisandrone: DESCRIPTORS_OUTPUT_FOLDER so it's clearly an output folder

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -0,0 +1,131 @@
...
+import BundleDescriptorService from './bundle-descriptor-service'
+import * as YAML from 'yaml'
+import * as fs from 'node:fs'
+import * as path from 'node:path'
+import { DESCRIPTORS_FOLDER } from '../paths'
+
+const DESCRIPTOR_VERSION = 'v4'
+const DESCRIPTOR_EXTENSION = '.yaml'
+const BUNDLE_DESCRIPTOR_NAME = 'descriptor' + DESCRIPTOR_EXTENSION
+const WIDGETS_DESCRIPTORS_FOLDER = path.join('ui', 'widgets')
```
cecchisandrone: can't we just use `widgets`?

cecchisandrone (approved):

PR #14, opened 2022-05-18
ENG-3595 Implemented Docker organization initialization

zonia3000:


src/commands/package.ts
```diff
@@ -0,0 +1,51 @@
+import { CliUx, Command, Flags } from '@oclif/core'
+import { BundleService } from '../services/bundle-service'
+import ConfigService, {
+  DOCKER_ORGANIZATION_PROPERTY
+} from '../services/config-service'
+
+export default class Package extends Command {
+  static description = 'Generates the Docker image for the bundle'
```
cecchisandrone: Generates the bundle Docker image

src/commands/package.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  static examples = [
+    '<%= config.bin %> <%= command.id %>',
+    '<%= config.bin %> <%= command.id %> --organization=my-org'
```
cecchisandrone: Maybe better to shorten to `--org`

src/commands/package.ts
```diff
@@ -0,0 +1,51 @@
...
+  configService = new ConfigService()
+
+  public async run(): Promise<void> {
+    BundleService.verifyBundleInitialized(process.cwd())
+    const { flags } = await this.parse(Package)
+    await this.getDockerOrganization(flags.organization)
+  }
+
+  private async getDockerOrganization(flagOrganization: string | undefined) {
+    if (flagOrganization) {
```
cecchisandrone: As agreed, overwrite current cached org

src/commands/package.ts
```diff
@@ -24,28 +24,41 @@ export default class Package extends Command {
...
   }
 
   private async getDockerOrganization(flagOrganization: string | undefined) {
+    const configuredOrganization = this.configService.getProperty(
+      DOCKER_ORGANIZATION_PROPERTY
+    )
+
     if (flagOrganization) {
+      if (configuredOrganization) {
+        this.configService.updateProperty(
```
cecchisandrone: at this point I would create an `addOrUpdateProperty` in the service

cecchisandrone (approved):

PR #15, opened 2022-05-19
ENG-3704 Implement api add command

raxkaynan:
The command adds an **internal** api claim to an mfe


src/commands/api/add.ts
```diff
@@ -0,0 +1,54 @@
+import { CliUx, Command, Flags } from '@oclif/core'
+import { ApiClaim, ApiType } from '../../models/api'
+import { ApiClaimService } from '../../services/api-claim-service'
+import { BundleService } from '../../services/bundle-service'
+
+export default class Add extends Command {
+  static description = 'Adds an internal API claim to the specified MFE component'
+
+  static examples = [
+    '$ entando-bundle-cli api add mfe1 ms1-api --serviceId ms1 --serviceUrl http://localhost:8080',
```
zonia3000: `$ entando-bundle-cli api add` can be replaced with placeholder `<%= config.bin %> <%= command.id %>`

src/services/mfe-config-service.ts
```diff
@@ -0,0 +1,32 @@
+import * as path from 'node:path'
+import * as fs from 'node:fs'
+import { MICROFRONTENDS_FOLDER } from '../paths'
+import { MfeConfig } from '../models/mfe-config'
+
+const MFE_CONFIG_FILE_NAME = 'mfe-config.json'
+const MFE_CONFIG_INDENTATION_SPACES = 4
```
zonia3000: Probably we should consider defining a global `INDENTATION_SPACES` constant for all the classes that use `JSON.stringify()`, since we have the same kind of definition in `BundleDescriptorService` and probably it will be needed somewhere else in the future.

src/commands/api/add.ts
```diff
@@ -0,0 +1,54 @@
...
+    {
+      name: 'claimName',
+      description: 'Name of the API claim',
+      required: true
+    }
+  ]
+
+  static flags = {
+    serviceId: Flags.string({
+      description: 'Micro Service ID within the Bundle',
```
cecchisandrone: 1. We agreed to call them _Microservice_ and _Micro Frontend_
2. Replace ID with name

src/commands/api/add.ts
```diff
@@ -0,0 +1,54 @@
...
+      required: true
+    }),
+    serviceUrl: Flags.string({
+      description: 'URL of the internal service',
```
cecchisandrone: Local microservice URL

src/commands/api/add.ts
```diff
@@ -0,0 +1,54 @@
...
+    const { args, flags } = await this.parse(Add)
+
+    const apiClaim: ApiClaim = {
+      name: args.claimName,
+      type: ApiType.Internal,
+      serviceId: flags.serviceId
+    }
+    const apiClaimService: ApiClaimService = new ApiClaimService()
+
+    CliUx.ux.action.start(`Adding a new API claim ${args.claimName} to ${args.mfeName}`)
```
cecchisandrone: Adding a new API claim named ${args.claimName} to Micro Frontend ${args.mfeName}

src/helpers/file-helper.ts
```diff
@@ -0,0 +1,12 @@
+import * as fs from 'node:fs'
```
cecchisandrone: Move this to FSService

src/models/api.ts
```diff
@@ -0,0 +1,11 @@
+export enum ApiType {
```
cecchisandrone: This should be in bundle-descriptor.ts

src/services/api-claim-service.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  constructor() {
+    this.bundleDescriptorService = new BundleDescriptorService(process.cwd())
+    this.mfeConfigService = new MfeConfigService()
+  }
+
+  public addInternalApiClaim(mfeName: string, apiClaim: ApiClaim, serviceUrl: string): void {
+    this.addApiClaim(mfeName, apiClaim)
+
+    this.updateMfeConfigApi(mfeName, apiClaim.name, serviceUrl)
```
cecchisandrone, line 24: If you add an invalid serviceUrl now it will result in adding the claim. Can you please verify and fix the test for this?

src/services/api-claim-service.ts
```diff
@@ -0,0 +1,80 @@
...
+    microfrontends[mfeIdx].apiClaims?.push(apiClaim)
+    const updatedBundleDescriptor: BundleDescriptor = {
+      ...bundleDescriptor,
+      microfrontends
+    }
+    this.bundleDescriptorService.writeBundleDescriptor(updatedBundleDescriptor)
+  }
+
+  private updateMfeConfigApi(mfeName: string, apiClaimName: string, url: string) {
+    if (!this.isValidUrl(url)) {
```
cecchisandrone: Validation should be done before adding the claim. Now if you add a bad service url it will add the claim to the descriptor but fails only after, this is wrong

src/services/api-claim-service.ts
raxkaynan: Oh yeah. Good catch

cecchisandrone (approved):

PR #16, opened 2022-05-20
ENG-3596 Added service for executing Docker build

zonia3000:


alepintus (approved):
Looks good

src/services/docker-service.ts
```diff
@@ -0,0 +1,39 @@
+import ProcessExecutorService from './process-executor-service'
+import debugFactory from './debug-factory-service'
+
+export type DockerBuildOptions = {
+  path: string
+  organization: string
+  name: string
+  version: string
```
cecchisandrone: better calling this tag (we are in docker context here)

src/services/process-executor-service.ts
```diff
@@ -0,0 +1,61 @@
+import { CLIError } from '@oclif/errors'
+import { spawn } from 'node:child_process'
+import { Writable } from 'node:stream'
+
+export type ProcessExecutionOptions = {
+  command: string
+  arguments?: string[]
+  /** Child process working directory */
+  cwd?: string
```
cecchisandrone: better workDir as cwd normally stands for current working directory. wdyt?

src/services/process-executor-service.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Writable where the child process output (both stdout and stderr) will be sent.
+   * This allows to send the output to the caller process stdout/stderr or redirect it to a file.
+   * If the field is not defined the output will be ignored.
+   */
+  outputStream?: Writable
```
cecchisandrone, line 14: isnt' maybe better to split stdout and stderr to allow fine grained config?

src/services/process-executor-service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export default class ProcessExecutorService {
+  /**
+   * Executes a long running child process and handles its output streams.
+   * @param options parameters for underlying spawn function and output configuration
+   * @returns a Promise that resolves when the process completes successfully or rejects when it fails
+   */
+  public static async executeProcess(
+    options: ProcessExecutionOptions
+  ): Promise<void> {
```
cecchisandrone, line 31: maybe useful to return exit code in the promise

src/services/process-executor-service.ts
```diff
@@ -0,0 +1,61 @@
...
+              `Command exited with code ${code} and signal ${signal}`
+            )
+          )
+        }
+
+        resolve()
+      })
+
+      process.on('error', function (error) {
+        reject(new CLIError(`Command failed due to error: ${error}`))
```
cecchisandrone: maybe better a generic `Error` here. Caller will catch and throw `CliError` if needed.

src/services/process-executor-service.ts
zonia3000: Not sure about this, maybe we need a more complex object, because if the process is killed a `signal` is returned and exit code is null.

zonia3000: Still having troubles with this.
One of the reasons of the Promise failure can be the `error` event, that is triggered for example if the command doesn't exist. Also in this case there is no exit code at all, since the process never started. Moreover, if I try to pass a number to the Promise `reject()` method the linter complains about `Expected the Promise rejection reason to be an Error`. So it looks that the only case when we should return the exit code is when it is equals to 0, that is not very useful.

cecchisandrone: Ok let's skip this for the moment

cecchisandrone (approved):

PR #17, opened 2022-05-20
ENG-3736 Added ms rm command

zonia3000:


test/helpers/temp-dir-helper.ts
```diff
@@ -0,0 +1,62 @@
+import * as fs from 'node:fs'
```
raxkaynan, line 1: This file's parent folder (`helpers`) might be confusing since it looks like it could contain test files as well (i.e. if I create a `src/helpers` folder, the files in `test/helpers` could be confused to correspond to test files of `src/helpers`). Maybe putting the test helpers in the root `test` directory is fine (or another alternative you could think of).

test/commands/ms/rm.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    })
+
+  test
+    .do(() => tempDirHelper.createInitializedBundleDir('test-bundle-no-ms'))
+    .command(['ms rm', 'not-existing-bundle'])
+    .catch(error => {
+      expect(error.message).to.contain('not found')
+    })
+    .it("Returns error if microservice to remove doesn't exist")
+})
```
raxkaynan, line 51: I think we should also add a test scenario where the command gets executed in a directory that isn't a bundle project, which would also return an error message.

raxkaynan (changes requested):

src/commands/ms/rm.ts
```diff
@@ -0,0 +1,26 @@
+import { CliUx, Command } from '@oclif/core'
+import { BundleService } from '../../services/bundle-service'
+import { MicroServiceService } from '../../services/microservice-service'
+
+export default class Rm extends Command {
+  static description =
+    'Removes a microservice component from the current bundle'
```
cecchisandrone, line 7: Can you also check the microservice word with other case or splitted and unify?

test/helpers/temp-dir-helper.ts
raxkaynan: As per discussion, resolving this.

src/services/microservice-service.ts
```diff
@@ -28,6 +28,25 @@ export class MicroServiceService {
...
+    const { microservices } = bundleDescriptor
+    const msIndex = microservices.findIndex(ms => ms.name === name)
+
+    if (msIndex === -1) {
+      throw new CLIError(`Microservice ${name} not found in Bundle descriptor`)
+    }
+
+    microservices.splice(msIndex, 1)
+
+    this.bundleDescriptorService.writeBundleDescriptor(bundleDescriptor)
```
cecchisandrone, line 47: Maybe better to remove the directory before updating the descriptor

cecchisandrone (changes requested):

cecchisandrone (approved):

PR #18, opened 2022-05-23
ENG-3735 amend command for removing microfrontend by given name

jeffgo10:


src/services/microfrontend-service.ts
```diff
@@ -31,6 +31,37 @@ export class MicroFrontendService {
...
+    const bundleDescriptor: BundleDescriptor =
+      this.bundleDescriptorService.getBundleDescriptor()
+    const { microfrontends } = bundleDescriptor
+
+    const updatedBundleDescriptor: BundleDescriptor = {
+      ...bundleDescriptor,
+      microfrontends: microfrontends.filter(({ name }) => name !== mfeName),
+    }
+
+    this.bundleDescriptorService.writeBundleDescriptor(updatedBundleDescriptor)
```
cecchisandrone: invert this in order to be sure folder is deleted before updating the desc

jeffgo10: done

cecchisandrone (approved):

PR #19, opened 2022-05-23
ENG-3746 add attribute in descriptor + rename bundle descriptor file

jeffgo10:


src/services/bundle-service.ts
```diff
@@ -14,7 +14,7 @@ export class BundleService {
   }
 
   private static descriptorExists(bundleDir: string): boolean {
-    const descriptorPath: string = path.resolve(bundleDir, 'bundle.json')
+    const descriptorPath: string = path.resolve(bundleDir, 'entando.json')
```
cecchisandrone: centralize this in path.ts, so you can reference also in tests

jeffgo10: done

cecchisandrone (approved):

PR #21, opened 2022-05-24
ENG-3706 Implement api add-ext command

raxkaynan:


src/commands/api/add-ext.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ApiType, ExternalApiClaim } from '../../models/bundle-descriptor'
+import { ApiClaimService } from '../../services/api-claim-service'
+import { BundleService } from '../../services/bundle-service'
+
+export default class AddExt extends Command {
+  static description =
+    'Adds an external API claim to the specified MFE component'
+
+  static examples = [
+    '$ <%= config.bin %> <%= command.id %> mfe1 ms1-api --bundleId my-bundle --serviceId ms1'
```
zonia3000: It looks that this `$` can be removed, indeed in the other commands we are not adding it and it is automatically added by Oclif.

src/services/api-claim-service.ts
```diff
@@ -24,11 +27,26 @@ export class ApiClaimService {
...
-
     this.updateMfeConfigApi(mfeName, apiClaim.name, serviceUrl)
   }
 
-  private addApiClaim(mfeName: string, apiClaim: ApiClaim): void {
+  public addExternalApiClaim(
+    mfeName: string,
+    apiClaim: ExternalApiClaim
+  ): void {
+    const url = this.cmService.getBundleMicroserviceUrl(
```
cecchisandrone, line 41: what if url is null or undefined (because bundleId and serviceId are invalid)? I guess connection issues will be handled by cmService instead by url will be null for sure

raxkaynan: I presumed that the api would respond with an error if bundleId and serviceId are invalid or non-existent, which would then cause the cli to throw an error as well (will be implemented in ENG-3708). If indeed it just returns a null value, I'll make the necessary changes to handle it.

cecchisandrone: I'm not sure about it, always better to check

src/services/api-claim-service.ts
```diff
@@ -24,11 +27,30 @@ export class ApiClaimService {
...
+      apiClaim.bundleId,
+      apiClaim.serviceId
+    )
+
+    if (!url) {
+      throw new CLIError('Failed to get microservice URL')
+    }
+
+    this.addApiClaim(mfeName, apiClaim)
```
cecchisandrone, line 50: Warning, we are adding an external API claim, so you don't need to check if the referenced MS is contained into the bundle (it's an external one, so the validation is done on CM side)

raxkaynan: Ah good one. Didn't notice that.

cecchisandrone (approved):

PR #22, opened 2022-05-24
ENG-3553 Build a single microservice

alepintus:


src/services/component-service.ts
```diff
@@ -39,6 +40,22 @@ export class ComponentService {
...
+    const allComponents = this.getAllComponents()
+    const component = allComponents.find(comp => comp.name === name)
+    if (component === undefined) {
+      throw new CLIError(`Component ${name} not found`)
+    }
+
+    return component
+  }
+
+  getAllComponents(): Component[] {
```
raxkaynan: `getComponents()`, without a specified type argument, already gets all components.

src/models/bundle-descriptor.ts
```diff
@@ -55,6 +55,7 @@ export type BundleDescriptor = {
   name: string
   /** Bundle version. It will be used as default Docker image tag */
   version: string
+  type: string
```
raxkaynan, line 58: Not sure why `type` has been added here.

src/commands/build.ts
```diff
@@ -0,0 +1,23 @@
...
+  static args = [
+    {
+      name: 'name',
+      description: 'The name of the component to build',
+      required: true
+    }
+  ]
+
+  public async run(): Promise<void> {
+    const { args } = await this.parse(Build)
```
raxkaynan, line 21: Lacks bundle verification (i.e. `BundleService.verifyBundleInitialized(process.cwd())`). If it gets added, there should be a corresponding test as well.

src/services/build-service.ts
```diff
@@ -0,0 +1,50 @@
+import * as path from 'node:path'
+import * as fs from 'node:fs'
+import { CLIError } from '@oclif/errors'
+import debugFactory from './debug-factory-service'
+
+import { MICROSERVICES_FOLDER } from '../paths'
+import ProcessExecutorService from './process-executor-service'
+import { ComponentService } from './component-service'
+
+export class BuildService {
```
raxkaynan: Do we need a separate `BuildService` class just for the build method/s?

@cecchisandrone What do you think about the `build()` method being in `ComponentService` instead?

raxkaynan (changes requested):

src/models/bundle-descriptor.ts
cecchisandrone: It's probably coming from the merge. We store which kind of descriptor it is (we renamed bundle.json into entando.json), in our case it will always be `bundle` but we may have other contexts

src/services/build-service.ts
cecchisandrone: yeah it can be a good idea, considering that there we have access to other services able to read project specific config. Also better using non-static methods if not for simple utility methods

src/services/build-service.ts
```diff
@@ -0,0 +1,50 @@
...
+    const componentService = new ComponentService()
+    const component = componentService.getComponent(name)
+
+    const { type, stack } = component
+
+    let componentPath: string
+    let buildCmd = ''
+    let buildCmdArgs: string[] = []
+
+    if (type === 'microservice' && stack === 'spring-boot') {
```
cecchisandrone: We agreed that this logic is not extensible and will be refactored when working on MFE build task. In my opinion there should be a `CommandGeneratorFactory` that instantiates the correct `CommandGenerator` for the given combination of stack, type and override (more on this [here](https://entando.myjetbrains.com/youtrack/issue/ENG-3671)). Also, the `CommandGenerator` can provide not only build but also `clean`, `package` and other commands. @zonia3000 how are you managing this for package part?

src/services/build-service.ts
```diff
@@ -0,0 +1,50 @@
...
+      componentPath = path.resolve(MICROSERVICES_FOLDER, name)
+      buildCmd = 'mvn'
+      buildCmdArgs = ['clean', 'test']
+    } else {
+      throw new CLIError(`${stack} ${type} build not implemented`)
+    }
+
+    BuildService.debug(
```
cecchisandrone: Merge in one line

src/commands/build.ts
```diff
@@ -0,0 +1,23 @@
...
+    CliUx.ux.action.start(`Building Component ${args.name}`)
```
cecchisandrone: component

src/services/build-service.ts
```diff
@@ -0,0 +1,50 @@
...
+      throw new CLIError(`Directory ${componentPath} not exists`)
+    }
+
+    process.chdir(componentPath)
+
+    return ProcessExecutorService.executeProcess({
+      command: buildCmd,
+      arguments: buildCmdArgs,
+      // Build output will be visible only in debug mode
+      outputStream: BuildService.debug.outputStream,
```
cecchisandrone: Build output (both stdout and stderr) in case of single component should be printed to stdout as reported in the ticket

src/services/component-service.ts
alepintus: getAllComponents removed

src/commands/build.ts
alepintus: verifyBundleInitialized added

src/services/build-service.ts
```diff
@@ -0,0 +1,50 @@
...
+    BuildService.debug(
+      `Building ${name} using ${buildCmd} ${buildCmdArgs.join(' ').trim()}`
+    )
+    BuildService.debug(`Component Path ${componentPath}`)
+
+    if (!fs.existsSync(componentPath)) {
+      throw new CLIError(`Directory ${componentPath} not exists`)
+    }
+
+    process.chdir(componentPath)
```
cecchisandrone: I think we should move this to the ProcessExecutorService. wdyt @zonia3000 ?

src/models/bundle-descriptor.ts
alepintus: The type is a new mandatory Field so I added it in some files where it was not present

cecchisandrone:
This is the current build failure output. This is not very clear for the user. Also let's avoid printing signal info if it's null
image

src/commands/build.ts
alepintus: done

src/services/build-service.ts
alepintus: done

src/services/build-service.ts
alepintus: Build moved to ComponentService

src/services/build-service.ts
alepintus: This is a change done by prettier, probably this is correct

src/services/build-service.ts
zonia3000: @cecchisandrone I can create a `CommandGeneratorFactory` in the package ticket if you want, the we can refactor the code.

src/services/build-service.ts
zonia3000: Options of the `executeProcess()` method accepts an optional parameter `workDir`, that is passed to the node `spawn` function. I think that should be the correct way to handle it, also considering when we will have to handle multiple processes in parallel, so each process may have a different working directory.

src/services/build-service.ts
cecchisandrone: Ok just sync with @alepintus

src/services/build-service.ts
cecchisandrone: I mean let's avoid double debug statement

zonia3000:
> This is the current build failure output. This is not very clear for the user. Also let's avoid printing signal info if it's null image

@cecchisandrone This will be fixed by my [PR](https://github.com/entando/entando-bundle-cli/pull/23/commits/1d01d7ac2ac559ceb14df7bf87a9170df441a564). I've refactored the class so that the promise returns an object of type `ProcessExecutionResult` and I'm using a custom exception class that handles the various cases.

src/services/build-service.ts
alepintus: Done

test/commands/build.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    stack: 'test',
+    type: ComponentType.MICROSERVICE
+  }
+  let executeProcessStub: sinon.SinonStub
+
+  afterEach(function () {
+    sinon.restore()
+  })
+
+  test
```
cecchisandrone, line 35: add a test for failed build

alepintus: done

cecchisandrone (approved):

PR #23, opened 2022-05-25
ENG-3701 Added parallel process executor

zonia3000:


src/services/fs-service.ts
```diff
@@ -19,7 +19,7 @@ export class FSService {
   }
 
   public checkBundleName(): void {
-    FSService.debug('checking bundle name if it\'s accepted')
+    FSService.debug("checking bundle name if it's accepted")
```
cecchisandrone: checking if bundle name is valid

src/services/process-executor-service.ts
```diff
@@ -1,5 +1,8 @@
 import { spawn } from 'node:child_process'
-import { Writable } from 'node:stream'
+import { EventEmitter, Writable } from 'node:stream'
+import { debugFactory } from './debug-factory-service'
+
+const PARALLELISM = 2
```
cecchisandrone: DEFAULT_PARALLEL_PROCESSES_SIZE

src/services/process-executor-service.ts
```diff
@@ -1,5 +1,8 @@
...
 
 export type ProcessExecutionOptions = {
   command: string
```
cecchisandrone, line 8: specify what is the default for workDir if not specified

src/services/process-executor-service.ts
```diff
@@ -20,45 +23,155 @@ export type ProcessExecutionOptions = {
...
-          reject(
-            new Error(`Command exited with code ${code} and signal ${signal}`)
-          )
+type QueuedProcessExecution = {
+  options: ProcessExecutionOptions
+  index: number
+}
+
+/**
+ * Executes multiple long running child processes and handles their output streams.
```
cecchisandrone, line 78: I would add a note regarding the execution order

src/services/process-executor-service.ts
```diff
@@ -20,45 +23,158 @@ export type ProcessExecutionOptions = {
...
+  constructor(
+    processesOptions: ProcessExecutionOptions[],
+    parallelism: number = DEFAULT_PARALLEL_PROCESSES_SIZE
+  ) {
+    super()
+    this.parallelism = parallelism
+    this.processesToWait = processesOptions.length
+    this.queue = processesOptions
+      .map((options, index) => ({ options, index }))
+      // reversing queue, so that when we use pop() the first process of the list
```
raxkaynan: `shift()` can be used to avoid having to reverse the array

src/services/process-executor-service.ts
```diff
@@ -20,45 +23,158 @@ export type ProcessExecutionOptions = {
...
-        cwd: options.workDir
-      })
+      const process = setUpProcess(options)
 
-      process.stdout.on('data', chunk => {
-        if (options.outputStream) {
-          options.outputStream.write(chunk)
+      process.on('exit', (code, signal) => {
+        if (code !== 0) {
+          reject(new ExecutionError(code === null ? signal! : code))
```
raxkaynan: You could also write `code ?? signal!` if you prefer it to be shorter :)
Same with [this line](https://github.com/entando/entando-bundle-cli/pull/23/files#diff-0d111d424524e7084e39844946aea6238e47d3198f4de8dbbf2fdaba1d361518R148)

raxkaynan (changes requested):
Just minor changes

cecchisandrone (approved):

src/services/process-executor-service.ts
zonia3000: thanks, much better!

PR #24, opened 2022-05-26
ENG-3762 implement aux requirements during initialization process

jeffgo10:


src/services/initializer-service.ts
```diff
@@ -54,6 +61,7 @@ export class InitializerService {
     fs.mkdirSync(this.filesys.getBundleFilePath('microservices'))
     fs.mkdirSync(this.filesys.getBundleFilePath('microfrontends'))
     fs.mkdirSync(this.filesys.getBundleFilePath('epc'))
+    fs.mkdirSync(this.filesys.getBundleFilePath('aux'))
```
zonia3000: You can use the constants defined in `paths` here

src/services/initializer-service.ts
```diff
@@ -85,6 +93,7 @@ export class InitializerService {
     this.filesys.createSubDirectoryIfNotExist('microfrontends')
     this.filesys.createSubDirectoryIfNotExist('.ent')
     this.filesys.createSubDirectoryIfNotExist('.ent', 'output')
+    this.filesys.createSubDirectoryIfNotExist('aux')
```
zonia3000: Same here

resources/aux/keycloak-template
```diff
@@ -0,0 +1,20 @@
+version: '3.1'
```
raxkaynan: Not sure why the template files aren't named with the `.yml` file extension.

test/services/fs-service.test.ts
```diff
@@ -73,6 +73,30 @@ describe('fs-service', () => {
       expect(fs.existsSync(filePath)).to.eq(true)
     })
 
+  test
+    .do(() => {
+      fs.mkdirSync(path.resolve(tempDirHelper.tmpDir, defaultBundleName))
+      fs.mkdirSync(path.resolve(tempDirHelper.tmpDir, defaultBundleName, 'aux'))
```
raxkaynan: This could be replaced with just one line:
`fs.mkdirSync(path.resolve(tempDirHelper.tmpDir, defaultBundleName, 'aux'), { recursive: true })`

src/services/fs-service.ts
```diff
@@ -51,14 +55,19 @@ export class FSService {
...
-      path.resolve(__dirname, '..', '..', RESOURCES_FOLDER, templateFileName)
+    let templateFileContent = fs.readFileSync(
+      path.resolve(__dirname, '..', '..', RESOURCES_FOLDER, templateFileName),
+      replaceDict ? 'utf8' : null,
     )
+    if (replaceDict) {
+      const placeholders: string[] = Object.keys(replaceDict)
+      templateFileContent = placeholders.reduce((currContent: string, placeholder: string) => (
+        currContent.replace(new RegExp(placeholder, 'g'), replaceDict[placeholder])
+      ), templateFileContent as string)
```
raxkaynan: I think another approach is:
```
for (placeholder of placeholders) {
  templateFileContent = templateFileContent.replace(new RegExp(placeholder, 'g'), replaceDict[placeholder])
}
```
So not sure if we need to remove the `no-array-reduce` rule.

test/services/fs-service.test.ts
```diff
@@ -73,6 +73,30 @@ describe('fs-service', () => {
...
+          path.join('aux', 'mysql-template'),
+          { '%BUNDLENAME%': defaultBundleName }
+        )
+        const filePath = path.resolve(
+          tempDirHelper.tmpDir,
+          defaultBundleName,
+          'aux',
+          'mysql.yml'
+        )
+        expect(fs.existsSync(filePath)).to.eq(true)
```
raxkaynan, line 115: Only the file existence is being asserted here. We should probably check if the generated file has the correct content as well.

raxkaynan (changes requested):

resources/aux/keycloak-template
cecchisandrone: yes, better to keep it for IDE readability

resources/aux/keycloak-template
```diff
@@ -0,0 +1,20 @@
...
+
+services:
```
cecchisandrone: for the moment I'll ignore those config as they need to be rechecked in other tickets

src/services/fs-service.ts
cecchisandrone: better, I think `reduce` is not needed here

src/services/fs-service.ts
```diff
@@ -7,6 +7,10 @@ import { RESOURCES_FOLDER } from '../paths'
 const ALLOWED_BUNDLE_NAME_REGEXP = /^[\w-]+$/
 const JSON_INDENTATION_SPACES = 4
 
+interface ContentReplaceDictionary {
```
cecchisandrone: just `TemplateVariables`

src/services/initializer-service.ts
```diff
@@ -112,6 +121,21 @@ export class InitializerService {
     this.filesys.createFileFromTemplate(['Dockerfile'], 'Dockerfile-template')
   }
 
+  public createDefaultAuxFiles(): void {
+    InitializerService.debug('creating auxfiles')
```
cecchisandrone: creating aux files

src/services/initializer-service.ts
```diff
@@ -112,6 +121,21 @@ export class InitializerService {
...
+
+    const replaceDict = { '%BUNDLENAME%': this.options.name }
+    const defaultYamls = ['mysql', 'postgresql', 'keycloak']
```
cecchisandrone: not needed, just read all files in the `aux` folder so we don't need to modify this if we add other aux files

test/commands/init.test.ts
```diff
@@ -149,14 +150,20 @@ describe('init', () => {
...
-  function checkFoldersStructure(bundleName: string) {
+  function checkFoldersStructure(bundleName: string, fromhub?: boolean) {
     checkBundleFile(bundleName, CONFIG_FOLDER)
     checkBundleFile(bundleName, CONFIG_FOLDER, CONFIG_FILE)
     checkBundleFile(bundleName, BUNDLE_DESCRIPTOR_FILE_NAME)
     checkBundleFile(bundleName, 'microservices')
     checkBundleFile(bundleName, 'microfrontends')
     checkBundleFile(bundleName, 'Dockerfile')
     checkBundleFile(bundleName, '.gitignore')
+    if (!fromhub) {
```
cecchisandrone: files should be there in any case, even if you get it from hub

src/services/initializer-service.ts
jeffgo10: done

src/services/initializer-service.ts
jeffgo10: done

resources/aux/keycloak-template
jeffgo10: done

test/services/fs-service.test.ts
jeffgo10: done

src/services/fs-service.ts
jeffgo10: done

test/services/fs-service.test.ts
jeffgo10: done

resources/aux/keycloak-template
jeffgo10: removed the extra configs (although i just copied those from examples out there)

src/services/fs-service.ts
jeffgo10: done

src/services/initializer-service.ts
jeffgo10: done

src/services/initializer-service.ts
jeffgo10: done

test/commands/init.test.ts
jeffgo10: done

.eslintrc
```diff
@@ -9,6 +9,7 @@
     "object-curly-spacing": "off",
     "comma-dangle": "off",
     "unicorn/prefer-module": "off",
+    "unicorn/no-array-reduce": "off",
```
raxkaynan: This is now unneeded

resources/aux/keycloak-yml-template
```diff
@@ -0,0 +1,9 @@
+version: '3.1'
```
raxkaynan: I believe it should be `keycloak-template.yml`, shouldn't it @cecchisandrone?

raxkaynan (changes requested):

resources/aux/keycloak-yml-template
jeffgo10: i was thinking about it. i just made it `default-keycloak.yml` since i saw how `default-config.json` was named. coming in next commit

.eslintrc
jeffgo10: done

resources/aux/keycloak-yml-template
jeffgo10: see file changes @raxkaynan

raxkaynan: I can see that there's also `Dockerfile-template` and `gitignore-template`. Quite inconsistent if some are named `default` and others `template`. What do you think @cecchisandrone?

jeffgo10: renamed every template prefix to `default`

cecchisandrone: ok

src/services/fs-service.ts
```diff
@@ -53,12 +57,24 @@ export class FSService {
 
   public createFileFromTemplate(
     pathSegments: string[],
-    templateFileName: string
+    templateFileName: string,
+    replaceDict?: TemplateVariables
```
cecchisandrone: rename param name to templateVariables

src/services/initializer-service.ts
```diff
@@ -49,11 +57,13 @@ export class InitializerService {
     const bundleDir = this.filesys.getBundleDirectory()
 
     fs.mkdirSync(bundleDir)
-    fs.mkdirSync(this.filesys.getBundleFilePath('.ent'))
-    fs.mkdirSync(this.filesys.getBundleFilePath('.ent', 'output'))
+    fs.mkdirSync(this.filesys.getBundleFilePath(CONFIG_FOLDER, 'output'), {
```
cecchisandrone: Refer to path.ts for all of these

src/services/initializer-service.ts
```diff
@@ -83,8 +93,9 @@ export class InitializerService {
   private async createDefaultDirectories() {
     this.filesys.createSubDirectoryIfNotExist('microservices')
```
cecchisandrone: same as above

src/services/initializer-service.ts
```diff
@@ -104,12 +115,32 @@ export class InitializerService {
...
+  public createDefaultAuxFiles(): void {
+    InitializerService.debug('creating aux files')
+
+    const replaceDict = { '%BUNDLENAME%': this.options.name }
+    const defaultYamls = fs
+      .readdirSync(
+        path.resolve(__dirname, '..', '..', RESOURCES_FOLDER, AUX_FOLDER)
+      )
+      .filter(filename => filename.slice(-3) === 'yml')
+      .map(filename => filename.slice(8))
```
cecchisandrone: safer replacing "default-" (as constant) with ""

src/services/initializer-service.ts
```diff
@@ -104,12 +115,32 @@ export class InitializerService {
...
   private createDockerfile() {
     InitializerService.debug('creating Dockerfile')
-    this.filesys.createFileFromTemplate(['Dockerfile'], 'Dockerfile-template')
+    this.filesys.createFileFromTemplate(['Dockerfile'], 'default-Dockerfile')
+  }
+
+  public createDefaultAuxFiles(): void {
+    InitializerService.debug('creating aux files')
+
+    const replaceDict = { '%BUNDLENAME%': this.options.name }
```
cecchisandrone: templateVariables

cecchisandrone (approved):

PR #26, opened 2022-05-27
ENG-3598 Package command implementation

zonia3000:
In addition to package command implementation, I refactored some parts of the code:

* I split the `Stack` enum in `MicroFrontendStack` and `MicroServiceStack` and I defined a `StackFor<T extends ComponentType>` type that is used to avoid that a MFE stack can be assigned to a MS and vice-versa. This is needed for correctly defining the `CommandsDefinition` type, which is used by the command factory to dynamically obtain the commands to build/package a component.
* I changed the promise of `ProcessExecutorService` to resolve always, as happens in the parallel version, so we can avoid to catch the exception in the command in order to display a proper error message
* I created a `BaseBuildCommand` that can be used by build, build-all and package commands, since it contains some methods for handling execution of commands on a set of components. I know that there is a lot of logic in the command, but we have to display logs and progress bars while the bundle is packed.


src/commands/base-build.ts
```diff
@@ -0,0 +1,135 @@
...
+import * as path from 'node:path'
+import { Component, ComponentType } from '../models/component'
+import {
+  MICROFRONTENDS_FOLDER,
+  MICROSERVICES_FOLDER,
+  OUTPUT_FOLDER
+} from '../paths'
+import { mkdirSync } from 'node:fs'
+
+export abstract class BaseBuildCommand extends Command {
```
cecchisandrone, line 23: This is appearing in the command help, it should be hidden

src/commands/build.ts
```diff
@@ -22,21 +22,18 @@ export default class Build extends Command {
...
-        } else {
-          console.error(
-            `Build failed, exit with code 1 and message ${error.message}`
-          )
-          this.exit(1)
-        }
+    const result = await componentService.build(args.name)
+    if (result !== 0) {
+      if (typeof result === 'number') {
+        console.error(`Build failed, exit with code ${result}`)
```
cecchisandrone: replace console.error() with this.error()

src/commands/base-build.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const executorService = new ParallelProcessExecutorService(executionOptions)
+
+    await this.parallelBuild(executorService, components)
+  }
+
+  private buildExecutionOptions(
+    components: Array<Component<ComponentType>>,
+    commandPhase: Phase
+  ) {
```
cecchisandrone: missing return type?

src/commands/base-build.ts
```diff
@@ -0,0 +1,135 @@
...
+import {
+  MICROFRONTENDS_FOLDER,
+  MICROSERVICES_FOLDER,
+  OUTPUT_FOLDER
+} from '../paths'
+import { mkdirSync } from 'node:fs'
+
+export abstract class BaseBuildCommand extends Command {
+  public async buildAllComponents(commandPhase: Phase): Promise<void> {
+    this.log('Building packages...')
```
cecchisandrone: Building components...

src/commands/build.ts
```diff
@@ -1,9 +1,9 @@
-import { CliUx, Command } from '@oclif/core'
+import { CliUx } from '@oclif/core'
 import { BundleService } from '../services/bundle-service'
 import { ComponentService } from '../services/component-service'
-import { ExecutionError } from '../services/process-executor-service'
+import { BaseBuildCommand } from './base-build'
 
-export default class Build extends Command {
+export default class Build extends BaseBuildCommand {
   static description = 'Build the component'
```
cecchisandrone: Build bundle components

src/commands/base-build.ts
```diff
@@ -0,0 +1,135 @@
...
+      progress.update(progress.value + 1)
+    })
+
+    const results = await executorService.execute()
+    progress.stop()
+
+    this.checkBuildResults(results, components)
+  }
+
+  public checkBuildResults(
```
cecchisandrone: should this be `private`?

src/services/microfrontend-service.ts
```diff
@@ -79,7 +86,7 @@ export class MicroFrontendService {
       throw new CLIError(`Directory ${mfedir} does not exist`)
     }
 
-    fs.rmSync(mfedir, { recursive: true, force: true });
+    fs.rmSync(mfedir, { recursive: true, force: true })
```
raxkaynan: I think it's safe to remove `force: true` here.

src/commands/package.ts
```diff
@@ -17,15 +28,47 @@ export default class Package extends Command {
...
+    BundleService.verifyBundleInitialized(bundleDir)
+
     const { flags } = await this.parse(Package)
-    await this.getDockerOrganization(flags.org)
+
+    const bundleDescriptorService = new BundleDescriptorService(bundleDir)
+    const bundleDescriptor = bundleDescriptorService.getBundleDescriptor()
+
+    const needsBuild = process.stdout.isTTY
+      ? await CliUx.ux.confirm('Rebuild components?')
```
cecchisandrone: Do you want to rebuild all bundle components?

src/commands/package.ts
```diff
@@ -17,15 +28,47 @@ export default class Package extends Command {
...
+
+    if (needsBuild) {
+      await this.buildAllComponents(Phase.Package)
+    }
+
+    const dockerOrganization = await this.getDockerOrganization(flags.org)
+
+    await this.buildMicroServicesDockerImages(
+      dockerOrganization,
+      bundleDescriptor.version
```
cecchisandrone: Docker images for MS will get their component version, not the bundle version

raxkaynan (changes requested):

src/commands/package.ts
```diff
@@ -17,15 +28,47 @@ export default class Package extends Command {
     org: Flags.string({
       char: 'o',
       description: 'Docker organization name'
+    }),
+    file: Flags.string({
+      char: 'f',
+      description: 'Name of the Dockerfile (default is Dockerfile)',
```
cecchisandrone: Bundle Dockerfile (default is Dockerfile)

src/commands/package.ts
cecchisandrone: Let's remove this as per the comment above

src/commands/package.ts
```diff
@@ -54,4 +97,75 @@ export default class Package extends Command {
     )
     return newOrganization
   }
+
+  private async buildMicroServicesDockerImages(
```
cecchisandrone: I would check for Dockerfile before calling the docker command and fail silently with it. In this way we can give a more explicit error like "Dockerfile not found for microservice <abc>. Please provide one in order to proceed with the bundle packaging"

src/commands/package.ts
```diff
@@ -54,4 +97,75 @@ export default class Package extends Command {
...
+    dockerOrganization: string,
+    tag: string
+  ) {
+    const componentService = new ComponentService()
+    const microServices = componentService.getComponents(
+      ComponentType.MICROSERVICE
+    )
+
+    this.log('Building microservices Docker images...')
```
cecchisandrone: Probably better to colorize macro steps or make it more visible (------ Building ... ----) in order to be more evident in the logs

src/commands/base-build.ts
```diff
@@ -0,0 +1,135 @@
...
+    results: ProcessExecutionResult[],
+    components: Array<Component<ComponentType>>
+  ): void {
+    if (results.some(result => result !== 0)) {
+      let errorMessage = 'Following components failed to build:\n'
```
cecchisandrone: The following components failed to build:\n

src/commands/package.ts
```diff
@@ -17,15 +28,47 @@ export default class Package extends Command {
     org: Flags.string({
       char: 'o',
       description: 'Docker organization name'
+    }),
+    file: Flags.string({
```
cecchisandrone: Let's modify a bit the UX. Add a flag -b to build all the components. If missing build will not be invoked

src/commands/package.ts
```diff
@@ -54,4 +97,75 @@ export default class Package extends Command {
...
+
+    await this.parallelBuild(executorService, microServices)
+  }
+
+  private async buildBundleDockerImage(
+    bundleDir: string,
+    bundleDescriptor: BundleDescriptor,
+    dockerOrganization: string
+  ) {
+    CliUx.ux.action.start('Creating bundle package')
```
cecchisandrone: Let's also log the full bundle image name at the end

cecchisandrone (approved):

PR #28, opened 2022-05-30
ENG-3763 Base mfe-config path on a microfrontend's publicFolder

raxkaynan:


test/commands/api/add.test.ts
```diff
@@ -131,7 +133,9 @@ describe('api add', () => {
 
   test
     .do(() => {
-      fs.rmSync(path.resolve('microfrontends', 'mfe1', 'mfe-config.json'))
+      fs.rmSync(
+        path.resolve('microfrontends', 'mfe1', 'public', 'mfe-config.json')
```
cecchisandrone: use constant for mfe-config.json

cecchisandrone (changes requested):

test/commands/api/add.test.ts
raxkaynan: that's just for the test. not sure if it's necessary to use a source constant there.

cecchisandrone: as we already have it defined probably worth to be used

raxkaynan: just changed `getMfeConfigPath()` from `private` to `public` instead so it can be used there

cecchisandrone (approved):

PR #30, opened 2022-05-31
ENG-3556 add build mfe command

alepintus:


src/services/command-factory-service.ts
```diff
@@ -30,7 +30,7 @@ const DEFAULT_COMMANDS: CommandsDefinition = {
   microservice: {
     'spring-boot': {
       clean: 'mvn clean',
-      build: 'mvn clean test',
+      build: 'mvn test',
```
cecchisandrone, line 33: @zonia3000 should we do the same for package?

src/services/component-service.ts
```diff
@@ -64,27 +67,30 @@ export class ComponentService {
     }))
   }
 
+  public static getComponentPath(component: Component<ComponentType>): string {
+    const { name, type } = component
+    const componentPath =
+      type === ComponentType.MICROSERVICE
```
cecchisandrone: I prefer using a switch here as we may have other components soon (EPC)

src/services/component-service.ts
```diff
@@ -102,6 +108,29 @@ export class ComponentService {
     return component
   }
 
+  validateComponent(component: Component<ComponentType>): void {
```
cecchisandrone, line 124: @alepintus @zonia3000 General bundle descriptor validation logic is being implemented in https://entando.myjetbrains.com/youtrack/issue/ENG-3788 so please add a TODO to refactor once that is done

src/services/command-factory-service.ts
zonia3000: I think that if we decide to remove the clean from build phase we can do the same for the package phase.
However I'm thinking that maybe we should invoke `mvn package` also in the build phase. Indeed, `mvn test` doesn't create the jar file and it would be a problem if a user executes `build` and then `pack` without the `-b` option.

src/services/component-service.ts
zonia3000: @alepintus Inside `base-build.ts` there is a similar logic (in `buildExecutionOptions()` method). We can use this new method also there. Thanks.

src/services/command-factory-service.ts
zonia3000: Update: we decided to remove the `-b` flag from `pack` command, so package phase always execute `mvn package` and it is correct to use `mvn test` here.

src/services/component-service.ts
```diff
@@ -64,27 +67,30 @@ export class ComponentService {
...
+  public static getComponentPath(component: Component<ComponentType>): string {
+    const { name, type } = component
+    const componentPath =
+      type === ComponentType.MICROSERVICE
+        ? path.resolve(MICROSERVICES_FOLDER, name)
+        : path.resolve(MICROFRONTENDS_FOLDER, name)
+    return componentPath
+  }
+
   public async build(name: string): Promise<ProcessExecutionResult> {
```
raxkaynan, line 86: Is it presumed that there would be no duplicate names for a micro frontend and microservice? @cecchisandrone

cecchisandrone: good point. @alepintus can you check this and enforce it at the moment of MFE/MS add command?

src/services/component-service.ts
alepintus: done

src/services/component-service.ts
alepintus: done

src/services/component-service.ts
alepintus: done

src/services/component-service.ts
```diff
@@ -99,6 +99,11 @@ export class ComponentService {
     })
   }
 
+  componentExists(name: string): boolean {
+    const component = this.getComponents().find(comp => comp.name === name)
```
raxkaynan: could just be `return this.getComponents().some(comp => comp.name === name)`

src/services/component-service.ts
```diff
@@ -113,6 +113,7 @@ export class ComponentService {
     return component
   }
 
+  // TODO Move the component validation to the Bundle validator
```
raxkaynan: The ticket number could be specified here to be clearer

raxkaynan (changes requested):

src/services/component-service.ts
alepintus: ticket number added

src/services/component-service.ts
alepintus: done

cecchisandrone (approved):

PR #32, opened 2022-06-01
ENG-3683 amend svc list command

jeffgo10:


src/commands/svc/list.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    const svcService: SvcService = new SvcService(process.cwd())
+    const services: string[] = flags.available
+      ? svcService.getAvailableServices()
+      : svcService.getActiveServices()
+    const serviceData = services.map(service => ({ service }))
+
+    const columns = {
+      service: {
+        header: flags.available ? 'Available Services' : 'Active Services'
```
cecchisandrone: Active Services -> Enabled Services

src/commands/svc/list.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
```
cecchisandrone, line 33: After listing the available services I would like to add a message like this: "Hint: you can add an auxiliary service just dropping a Docker Compose file in the svc folder (e.g. svc/my-service.yml)"

src/services/svc-service.ts
```diff
@@ -29,10 +29,20 @@ export class SvcService {
       .map(filename => filename.slice(0, -4))
```
cecchisandrone, line 29: Always better to filter specifically for files ending with .yml

jeffgo10: @cecchisandrone there's a filter function above that exactly does what you mentioned. the `map` only extracts the filenames, hence removing extension types.

src/commands/svc/list.ts
jeffgo10: done

src/commands/svc/list.ts
jeffgo10: log added at the end of table listing

cecchisandrone (approved):

PR #33, opened 2022-06-02
ENG-3785 Add output directory cleanup to ms and mfe remove operations

raxkaynan:
This PR also renames every [Mm]icro**S**ervice occurrence as [Mm]icro**s**ervice


src/commands/ms/add.ts
```diff
@@ -15,8 +15,8 @@ export default class Add extends Command {
   static flags = {
     stack: Flags.string({
       description: 'Microservice stack',
-      options: [MicroServiceStack.SpringBoot, MicroServiceStack.Node],
-      default: MicroServiceStack.SpringBoot
+      options: [MicroserviceStack.SpringBoot, MicroserviceStack.Node],
```
zonia3000: I would replace this with `Object.values(MicroserviceStack)`, so we are dynamic (and the same in MFE add command).

src/commands/ms/add.ts
```diff
@@ -1,8 +1,8 @@
 import { CliUx, Command, Flags } from '@oclif/core'
-import { MicroService } from '../../models/bundle-descriptor'
-import { MicroServiceStack } from '../../models/component'
+import { Microservice } from '../../models/bundle-descriptor'
```
cecchisandrone, line 2: @raxkaynan I think we agreed to leave `MicroService` in the code. Is there a reason for this?

cecchisandrone: Ok we agreed on renaming as found on Wikipedia and other official sources

cecchisandrone (approved):

PR #36, opened 2022-06-06
ENG-3788 Implemented bundle descriptor validation

zonia3000:
I wanted to create a validation mechanism that takes advantage of TypeScript features to ensure that, if we change the `BundleDescriptor` definition, the validation code will not compile until it matches the new structure. So I defined some types that represent validation constraints over other (generic) types.

The object holding `BundleDescriptor` constraints is a constant defined in the following way:

    const BUNDLE_DESCRIPTOR_CONSTRAINTS: ObjectConstraints<BundleDescriptor>

Properties like `required`, `type` and `isArray` are enforced by TypeScript to assume the value of the binding type, then it is possible to add some custom `Validator`s for things like checking RegExp or validate enum values.

Type definitions may look a bit complicated, due to the recursive nature of the problem and the various cases to handle. If you think that this causes too much headache we might consider alternative approaches, like JSON Schema validation.
It looks that there is [a library](https://www.npmjs.com/package/typescript-json-schema) able to generate JSON Schema from TypeScript types. Probably that is a more standard way to work, but implies to rely on external libraries.

I think that we should also discuss which fields should really be mandatory. Indeed with current setup the Jeff's bundle we are using for testing is not valid, because it misses the following required properties:

* `$.microservices[0].dbms`
* `$.microfrontends[0].titles`
* `$.microfrontends[0].publicFolder`
* `$.microfrontends[0].customUiPath`
* `$.microfrontends[0].group`

After introducing the validation some tests failed because they were stubbing the bundle descriptor with invalid objects. For some of them I stubbed the validation, while others have been fixed providing fake components having valid structure. I added a `ComponentHelper` class for this.


src/services/bundle-service.ts
```diff
@@ -12,10 +15,55 @@ export class BundleService {
...
+    const descriptorPath = path.resolve(bundleDir, BUNDLE_DESCRIPTOR_FILE_NAME)
+    const descriptorFileContent = fs.readFileSync(descriptorPath, 'utf-8')
+    try {
+      const parsedDescriptor: any = JSON.parse(descriptorFileContent)
+      const bundleDescriptor =
+        ConstraintsValidatorService.validateObjectConstraints(
+          parsedDescriptor,
+          BUNDLE_DESCRIPTOR_CONSTRAINTS
+        )
+      BundleService.checkDuplicatedComponentNames(bundleDescriptor)
```
cecchisandrone: we also have the same validation in addMicroFrontend/addMicroservice methods. Replace that with this

src/services/bundle-service.ts
```diff
@@ -1,7 +1,10 @@
 import { CLIError } from '@oclif/errors'
 import * as fs from 'node:fs'
 import * as path from 'node:path'
+import { BundleDescriptor } from '../models/bundle-descriptor'
+import { BUNDLE_DESCRIPTOR_CONSTRAINTS } from '../models/bundle-descriptor-constraints'
 import { BUNDLE_DESCRIPTOR_FILE_NAME } from '../paths'
+import { ConstraintsValidatorService } from './constraints-validator-service'
 
 export class BundleService {
   public static isBundleInitialized(bundleDir: string): boolean {
```
cecchisandrone: Remove this method, I think is not relevant

src/services/bundle-service.ts
```diff
@@ -12,10 +15,55 @@ export class BundleService {
     if (!BundleService.isBundleInitialized(bundleDir)) {
```
cecchisandrone: verifyBundleInitialized -> isValidBundleProject

src/services/bundle-service.ts
```diff
@@ -12,10 +15,55 @@ export class BundleService {
...
   private static descriptorExists(bundleDir: string): boolean {
-    const descriptorPath: string = path.resolve(bundleDir, BUNDLE_DESCRIPTOR_FILE_NAME)
+    const descriptorPath: string = path.resolve(
+      bundleDir,
+      BUNDLE_DESCRIPTOR_FILE_NAME
+    )
     return fs.existsSync(descriptorPath)
   }
+
+  private static verifyBundleDescriptorStructure(bundleDir: string): void {
```
cecchisandrone: 1. validateBundleDescriptor
2. Shouldn't it be in bundle-descriptor-service?

src/services/bundle-service.ts
```diff
@@ -12,10 +15,55 @@ export class BundleService {
     if (!BundleService.isBundleInitialized(bundleDir)) {
       throw new CLIError(`${bundleDir} is not an initialized Bundle project`)
     }
+
+    BundleService.verifyBundleDescriptorStructure(bundleDir)
   }
 
   private static descriptorExists(bundleDir: string): boolean {
-    const descriptorPath: string = path.resolve(bundleDir, BUNDLE_DESCRIPTOR_FILE_NAME)
+    const descriptorPath: string = path.resolve(
```
cecchisandrone: I think we don't need this anymore as presence of the descriptor is needed in  validateBundleDescriptor

src/models/bundle-descriptor-constraints.ts
```diff
@@ -0,0 +1,288 @@
...
+  titles: {
+    required: true,
+    validators: [isMapOfStrings],
+    items: {}
+  },
+  publicFolder: {
+    required: true,
+    type: 'string'
+  },
+  customUiPath: {
```
cecchisandrone: this should be removed, also from descriptor

src/models/bundle-descriptor-constraints.ts
```diff
@@ -0,0 +1,288 @@
...
+    required: true,
+    type: 'string'
+  },
+  configUi: {
```
cecchisandrone: this should be removed, also from descriptor

src/services/bundle-service.ts
```diff
@@ -12,10 +15,55 @@ export class BundleService {
...
+    } catch (error) {
+      throw new CLIError(
+        BUNDLE_DESCRIPTOR_FILE_NAME +
+          ' is not valid.\n' +
+          (error as Error).message
+      )
+    }
+  }
+
+  private static checkDuplicatedComponentNames(
```
cecchisandrone: same as above

src/services/constraints-validator-service.ts
```diff
@@ -0,0 +1,252 @@
...
+type BaseArrayConstraints<T, K extends keyof T> = {
+  isArray: true
+  required: IsRequired<T, K>
+}
+type PrimitiveArrayConstraints<T, K extends keyof T> = BaseArrayConstraints<
+  T,
+  K
+> &
+  PrimitiveConstraints<T, K>
+type ObjectArrayConstraints<T, K extends keyof T> = BaseArrayConstraints<
```
cecchisandrone: very strange formatting, can we enhance this?

src/models/bundle-descriptor-constraints.ts
```diff
@@ -0,0 +1,288 @@
...
+    required: true,
+    type: 'string',
+    validators: [values(MicroServiceStack)]
+  },
+  deploymentBaseName: {
+    required: false,
+    type: 'string'
+  },
+  dbms: {
+    required: true,
```
cecchisandrone: not required and it should be an enumeration (check with Luca the values)

src/models/bundle-descriptor-constraints.ts
```diff
@@ -0,0 +1,288 @@
...
+    type: 'string',
+    validators: [values(MicroFrontendStack)]
+  },
+  titles: {
+    required: true,
+    validators: [isMapOfStrings],
+    items: {}
+  },
+  publicFolder: {
+    required: true,
```
cecchisandrone: I'm not sure this is required as we should have a default value for it. @raxkaynan ?

src/models/bundle-descriptor-constraints.ts
```diff
@@ -0,0 +1,288 @@
...
+
+export const ALLOWED_NAME_REGEXP = /^[\w-]+$/
+export const INVALID_NAME_MESSAGE =
+  'Only alphanumeric characters, underscore and dash are allowed'
+
+// Validators
+
+const nameRegExpValidator = regexp(ALLOWED_NAME_REGEXP, INVALID_NAME_MESSAGE)
+
+function regexp(regexp: RegExp, message: string): Validator {
```
cecchisandrone: shouldn't those methods be in constraint-validator-service.ts?

src/services/bundle-service.ts
zonia3000: Do you mean that we should move the check in `BundleDescriptorService.validateBundleDescriptor()` or that we should let the CLI failing with `ENOENT: no such file or directory entando.json`?

src/services/bundle-service.ts
zonia3000: That uses `componentService.componentExists(newComponentName)`, it doesn't loop on all the components searching for duplicates. The logic is different.

src/services/bundle-service.ts
cecchisandrone: Just move the check

src/services/constraints-validator-service.ts
zonia3000: This issue depends on the `printWidth` setting in `.prettierrc`. Setting it to 90 instead of 80 fixed the bad formatting.
However I noticed that it is possible to force prettier to keep a newline after the `=` of a type definition adding a `|` to it.
It automatically does that for some union types, so I preferred that approach.

src/models/bundle-descriptor-constraints.ts
zonia3000: I made it optional and used the default value when it is undefined

cecchisandrone (approved):

PR #37, opened 2022-06-06
ENG-3790 Added hook for customizing CLI name

zonia3000:


alepintus (approved):
Looks good

cecchisandrone:
@zonia3000 I tested and it doesn't work with /dev/run. I also checked after `npm pack`

zonia3000:
> @zonia3000 I tested and it doesn't work with /dev/run. I also checked after `npm pack`

Should be fixed now. It looks that it is a bug of the `oclif generate hook` command. It automatically updates the `package.json` file with the path of the hook but it uses the `src` folder instead of `dist`.

cecchisandrone (approved):

PR #39, opened 2022-06-07
ENG-3689 amend svc stop command

jeffgo10:


src/services/svc-service.ts
```diff
@@ -83,6 +83,48 @@ export class SvcService {
...
+      .join(' ')} up --build -d`
+
+    try {
+      ProcessExecutorService.executeProcess({
+        command: cmd,
+        outputStream: process.stdout,
+        errorStream: process.stdout,
+        workDir: this.parentDirectory
+      })
+    } catch (error) {
```
zonia3000: I've just seen this and I noticed that the comment that describe the `executeProcess()` function is outdated. It states "returns a Promise that resolves when the process completes successfully or rejects when it fails", however after one of last refactorings, we made the promise always resolving successfully, returning a `ProcessExecutionResult`. This should be propagated to the command and the error should be handled there, like done in `build.ts`. In that case, when the child process fail with an exit code, that exit code is used to exit the caller process. Probably it could be useful to introduce a function that can be shared by multiple commands.

jeffgo10: hi @zonia3000 thank you for your suggestion. I have changed how the command level handles the process execution instances. please take a look and review and let me know your feedback. thanks!

src/commands/svc/stop.ts
```diff
@@ -0,0 +1,49 @@
+import { CliUx, Flags } from '@oclif/core'
+import { BundleService } from '../../services/bundle-service'
+import { SvcService } from '../../services/svc-service'
+import { SvcProcessResult, ServiceTypes } from './svc-process'
+
+export default class Stop extends SvcProcessResult {
+  static strict = false
+  static description = 'Stop running enabled auxiliary services'
```
zonia3000: Stops

src/commands/svc/svc-process.ts
```diff
@@ -0,0 +1,44 @@
...
+export enum ServiceTypes {
+  START = 'Starting',
+  STOP = 'Stopping'
+}
+
+export abstract class SvcProcessResult extends Command {
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+  protected checkResult(
+    result: any,
+    serviceType: string,
```
zonia3000: Here you can use more specific types: `ProcessExecutionResult` and `ServiceTypes`

src/services/svc-service.ts
```diff
@@ -82,12 +91,44 @@ export class SvcService {
...
+  public stopServices(services: string[]): Promise<ProcessExecutionResult> {
+    this.precheckEnabledServices(services)
+
+    SvcService.debug(`stopping service ${services.join(', ')}`)
+
+    return this.processDockerExecution(SERVICE_TYPE.DOCKER_STOP, services)
+  }
+
+  private processDockerExecution(
+    serviceType: string,
```
zonia3000: Here you can use more specific type: `SERVICE_TYPE`. What about renaming the enum as `ServiceType`? All the other enums we have defined are using camel case. Maybe also `DockerServiceType`, since we have a similar enum in `svc-process.ts`

src/services/svc-service.ts
zonia3000: Thanks @jeffgo10. For me it looks ok now. I've added just some minor comments.
Could you please also update the outdated comment above `executeProcess()` function stating that the promise always resolves?

jeffgo10: sure @zonia3000 will do

src/commands/svc/stop.ts
jeffgo10: I beg to differ. `Stop` is the right word to go with `services` since the subject `services` is a plural

src/services/svc-service.ts
jeffgo10: done @zonia3000 pls check if I described it right

src/commands/svc/svc-process.ts
jeffgo10: done

src/services/svc-service.ts
jeffgo10: done. thanks for pointing it out

src/commands/svc/stop.ts
zonia3000: I wrote it because all the other command descriptions are using the third-person singular verb. In the sense "This command does this thing".

jeffgo10: I misunderstood it then. I'll update the other descriptions on other commands as well. They are using the same grammer

jeffgo10: done

src/commands/svc/enable.ts
```diff
@@ -3,7 +3,7 @@ import { BundleService } from '../../services/bundle-service'
 import { SvcService } from '../../services/svc-service'
 
 export default class Enable extends Command {
-  static description = 'Enable auxiliary services'
+  static description = 'Enables auxiliary services'
```
cecchisandrone: As agreed, let's use imperative verbs

src/commands/svc/stop.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  static examples = [
+    '<%= config.bin %> <%= command.id %> --all',
+    '<%= config.bin %> <%= command.id %> ext-service',
+    '<%= config.bin %> <%= command.id %> ext-service1 ext-service2'
+  ]
+
+  static flags = {
+    all: Flags.boolean({
+      description: 'Stops all running enabled services in the bundle descriptor'
```
cecchisandrone: "Stops all enabled auxiliary services" is enough

src/commands/svc/start.ts
```diff
@@ -1,11 +1,11 @@
...
+import { SvcProcessResult, ServiceTypes } from './svc-process'
 
-export default class Start extends Command {
+export default class Start extends SvcProcessResult {
   static strict = false
-  static description = 'Start enabled auxiliary services'
+  static description = 'Starts enabled auxiliary services'
 
   static examples = [
     '<%= config.bin %> <%= command.id %> --all',
```
cecchisandrone: 'Start all enabled auxiliary services' is enough

src/commands/svc/stop.ts
```diff
@@ -0,0 +1,49 @@
+import { CliUx, Flags } from '@oclif/core'
+import { BundleService } from '../../services/bundle-service'
+import { SvcService } from '../../services/svc-service'
+import { SvcProcessResult, ServiceTypes } from './svc-process'
+
+export default class Stop extends SvcProcessResult {
+  static strict = false
+  static description = 'Stops running enabled auxiliary services'
```
cecchisandrone: Stops enabled auxiliary services

src/commands/svc/svc-process.ts
```diff
@@ -0,0 +1,47 @@
+import { Command } from '@oclif/core'
+import { ProcessExecutionResult } from '../../services/process-executor-service'
+
+export enum ServiceTypes {
+  START = 'Starting',
+  STOP = 'Stopping'
+}
+
+export abstract class SvcProcessResult extends Command {
```
cecchisandrone: rename file to svc-process-result.ts

src/commands/svc/enable.ts
zonia3000: Sorry @jeffgo10 for making you made the wrong changes. @cecchisandrone I missed that we agreed that format. We should also change the description of other commands (like init, pack, list, ...) since they use the other form.

cecchisandrone: My fault, it was like that before but I checked other CLIs and they use imperative form

jeffgo10: no worries @zonia3000, I already changed the description form in other commands like the ones you mentioned

src/commands/svc/stop.ts
jeffgo10: done

src/commands/svc/start.ts
jeffgo10: done

cecchisandrone (approved):

src/commands/svc/stop.ts
jeffgo10: done

src/commands/svc/svc-process.ts
jeffgo10: done

PR #44, opened 2022-06-08
ENG-3557 add build --all-mfe command

alepintus:


src/commands/build.ts
```diff
@@ -19,16 +23,33 @@ export default class Build extends BaseBuildCommand {
...
+    if (Object.keys(flags).length > 0 && args.name !== undefined) {
+      this.error(
+        `Build failed, please use one flag or write the component name as argument`,
+        { exit: 1 }
+      )
+    }
+
+    if (flags['all-mfe']) {
+      CliUx.ux.action.start(`Building all microfrontends`)
+      await this.buildAllComponents(Phase.Build, ComponentType.MICROFRONTEND)
```
zonia3000, line 50: I don't think we should use the spinner together with the progress bar. It causes the progress bar to split during the update, like this:
```
progress [----------------------------------------] 0% | ETA: 0s | 0/2
progress [========================================] 100% | ETA: 0s | 2/2
```
I noticed this using the `build --all-ms` command

alepintus: spinner removed

zonia3000 (approved):

cecchisandrone (approved):

PR #46, opened 2022-06-09
ENG-3791 Implement micro frontend types

raxkaynan:
* adds --type and --slot flags to mfe add command
* adds version field to yaml bundle descriptor
* redos union type constraints
* refactors tests


src/commands/mfe/add.ts
```diff
@@ -17,6 +21,15 @@ export default class Add extends Command {
...
       options: Object.values(MicroFrontendStack),
       default: MicroFrontendStack.React
+    }),
+    type: Flags.string({
+      description: 'Micro Frontend type',
+      options: Object.values(MicroFrontendType),
+      default: MicroFrontendType.Widget
+    }),
+    slot: Flags.string({
+      description: 'Micro Frontend app-builder slot',
```
cecchisandrone: Micro Frontend App Builder slot (only if type=app-builder)

src/commands/mfe/add.ts
```diff
@@ -33,9 +46,15 @@ export default class Add extends Command {
 
     const { args, flags } = await this.parse(Add)
 
+    if (flags.slot && flags.type !== MicroFrontendType.AppBuilder) {
```
cecchisandrone, line 49: Use dependsOn instead https://oclif.io/docs/flags

raxkaynan: `dependsOn` just checks if the other flag exists, not allowing to check for specific value. That's actually an [open issue](https://github.com/oclif/oclif/issues/323) in oclif.

cecchisandrone (approved):

PR #50, opened 2022-06-10
ENG-3815 Added method for checking Docker images existence

zonia3000:


src/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { Writable } from 'node:stream'
+
+export class InMemoryWritable extends Writable {
```
cecchisandrone, line 3: missing test?

src/commands/publish.ts
```diff
@@ -0,0 +1,40 @@
...
+        ))
+      ) {
+        console.warn(
+          'One or more Docker images are missing. Running pack command.'
+        )
+        // TODO: ENG-3826
+      }
+    } else {
+      console.warn(
+        'No configured Docker organization found. Running pack command.'
```
cecchisandrone: if just org is missing why do we need to run pack? isn't enough to specify flag -o to publish? I don't remember what we decided about org in publish

src/services/docker-service.ts
```diff
@@ -113,4 +121,76 @@ export class DockerService {
...
+    bundleDescriptor: BundleDescriptor,
+    organization: string
+  ): Promise<boolean> {
+    const images = DockerService.getBundleDockerImages(
+      bundleDescriptor,
+      organization
+    )
+
+    // Listing all the expected images
+    let command = 'docker image ls '
```
cecchisandrone: I think better to extract docker command in a constant as it can be easier to refactor

src/commands/publish.ts
```diff
@@ -0,0 +1,40 @@
+import { Command } from '@oclif/core'
+import { BundleDescriptorService } from '../services/bundle-descriptor-service'
+import {
+  ConfigService,
+  DOCKER_ORGANIZATION_PROPERTY
+} from '../services/config-service'
+import { DockerService } from '../services/docker-service'
+
+export default class Publish extends Command {
+  static description = 'Publish bundle Docker images'
```
cecchisandrone, line 10: image
I would rename image -> images in pack command as well

src/commands/publish.ts
zonia3000: Would a message like this be ok? `No configured Docker organization found. Please run the command with --org flag`?

cecchisandrone: yes perfect

cecchisandrone (approved):

PR #53, opened 2022-06-14
ENG-3673 modified mysql yaml default configuration

jeffgo10:
configuration is based from default yaml configuration generated from generator-jhipster-entando

### DIFF
--- a/resources/svc/default-mysql.yml
+++ b/resources/svc/default-mysql.yml
@@ -8,7 +8,7 @@ services:
     # volumes:
     #   - ~/volumes/entando-bundle-cli/%BUNDLENAME%/mysql/:/var/lib/mysql/
     environment:
-      - MYSQL_RANDOM_ROOT_PASSWORD=true
+      - MYSQL_ROOT_PASSWORD=%BUNDLENAME%-password
       - MYSQL_USER=%BUNDLENAME%-user
       - MYSQL_PASSWORD=%BUNDLENAME%-password
       - MYSQL_DATABASE=%BUNDLENAME%
--- a/resources/svc/default-mysql.yml
+++ b/resources/svc/default-mysql.yml
@@ -1,5 +1,5 @@
 # This configuration is intended for development purpose, it's **your** responsibility to harden it for production
-version: '3.8'
+version: '3.7'
 
 services:
   mysql:
--- a/resources/svc/default-mysql.yml
+++ b/resources/svc/default-mysql.yml
@@ -1,18 +1,16 @@
-version: '3.1'
+# This configuration is intended for development purpose, it's **your** responsibility to harden it for production
+version: '3.8'
 
 services:
   mysql:
-    image: mysql
-    restart: always
+    image: mysql:8.0.26
+    # volumes:
+    #   - ~/volumes/entando-bundle-cli/%BUNDLENAME%/mysql/:/var/lib/mysql/
     environment:
-      - MYSQL_DATABASE=%BUNDLENAME%-db
-      - MYSQL_ROOT_PASSWORD=adminadmin
+      - MYSQL_ALLOW_EMPTY_PASSWORD=yes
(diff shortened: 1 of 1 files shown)


resources/svc/default-mysql.yml
```diff
@@ -1,18 +1,16 @@
-version: '3.1'
+# This configuration is intended for development purpose, it's **your** responsibility to harden it for production
+version: '3.8'
 
 services:
   mysql:
-    image: mysql
-    restart: always
+    image: mysql:8.0.26
```
alepintus, line 6: Hi @jeffgo10 , The image version could be moved to package.json and added here as variable. @cecchisandrone WDYT?

resources/svc/default-mysql.yml
```diff
@@ -1,18 +1,16 @@
-version: '3.1'
+# This configuration is intended for development purpose, it's **your** responsibility to harden it for production
+version: '3.8'
```
cecchisandrone: let's use 3.7 as 3.8 is too recent

resources/svc/default-mysql.yml
```diff
@@ -1,18 +1,16 @@
...
 
 services:
   mysql:
-    image: mysql
-    restart: always
+    image: mysql:8.0.26
+    # volumes:
+    #   - ~/volumes/entando-bundle-cli/%BUNDLENAME%/mysql/:/var/lib/mysql/
```
cecchisandrone, line 9: what's this?

alepintus (changes requested):

resources/svc/default-mysql.yml
jeffgo10: @cecchisandrone this is an option to externalize your data, but I just copied that exact comment from the generated yaml from `generator-jhipster-entando`. Could be a quick option on externalizing data. do you want to remove this?

resources/svc/default-mysql.yml
cecchisandrone: We may do this to have the same approach as keycloak. In any case this is not a custom image.

jeffgo10: @alepintus i'm up for it. should we implement this @cecchisandrone ?

resources/svc/default-mysql.yml
cecchisandrone: If you want to keep this, put a comment explaining it

resources/svc/default-mysql.yml
cecchisandrone: @jeff keep it as it is. @alepintus will do the version centralization in the scope of ENG-3672

resources/svc/default-mysql.yml
jeffgo10: done

resources/svc/default-mysql.yml
jeffgo10: done

resources/svc/default-mysql.yml
```diff
@@ -1,18 +1,17 @@
...
-    image: mysql
-    restart: always
+    image: mysql:8.0.26
+    # If you want to relocate the mysql data (volume) to your local directory, uncomment the lines below
+    # volumes:
+    #   - ~/volumes/entando-bundle-cli/%BUNDLENAME%/mysql/:/var/lib/mysql/
     environment:
-      - MYSQL_DATABASE=%BUNDLENAME%-db
-      - MYSQL_ROOT_PASSWORD=adminadmin
+      - MYSQL_ALLOW_EMPTY_PASSWORD=yes
```
cecchisandrone: Empty password? Let's use `%BUNDLENAME%-user` and `%BUNDLENAME%-password`, same for postgres

jeffgo10: done. I added MYSQL_RANDOM_ROOT_PASSWORD as root password is needed in order the service to work

cecchisandrone (approved):

PR #57, opened 2022-06-15
ENG-3672 add keycloak service

alepintus:


src/services/initializer-service.ts
```diff
@@ -100,6 +110,9 @@ export class InitializerService {
     this.filesys.createSubDirectoryIfNotExist(CONFIG_FOLDER)
     this.filesys.createSubDirectoryIfNotExist(...OUTPUT_FOLDER)
     this.filesys.createSubDirectoryIfNotExist(SVC_FOLDER)
+    this.filesys.createSubDirectoryIfNotExist(...KEYCLOAK_FOLDER)
```
cecchisandrone, line 117: better creating a single function to group keycloak folders creation

src/services/initializer-service.ts
```diff
@@ -147,6 +171,33 @@ export class InitializerService {
         templateVariables
       )
     }
+
+    const keycloakRealm = path.join(
```
cecchisandrone, line 183: group keycloak stuff in a function for better readability

package.json
```diff
@@ -104,5 +104,10 @@
     "ts-node": "10.7.0",
     "tslib": "2.3.1",
     "typescript": "4.6.3"
+  },
+  "org_entando_dependencies": {
+    "mysql": "mysql:8.0.26",
```
cecchisandrone: put here only the version not the full image name

alepintus: done

src/services/initializer-service.ts
alepintus: done

src/services/initializer-service.ts
alepintus: done

cecchisandrone (approved):

PR #58, opened 2022-06-16
ENG-3829 Add name to and remove code from yaml descriptors

raxkaynan:
Also changes descriptor versions to "v5"

### DIFF
--- a/src/models/yaml-bundle-descriptor.ts
+++ b/src/models/yaml-bundle-descriptor.ts
@@ -23,7 +23,7 @@ export type YamlWidgetDescriptor = {
 
 export type YamlPluginDescriptor = {
   name: string
-  descriptorVersion: 'v5'
+  descriptorVersion: string
   image: string
   deploymentBaseName?: string
   dbms: DBMS
--- a/src/models/yaml-bundle-descriptor.ts
+++ b/src/models/yaml-bundle-descriptor.ts
@@ -13,7 +13,7 @@ export type YamlWidgetDescriptor = {
   name: string
   titles: { [lang: string]: string }
   group: string
-  version: string
+  descriptorVersion: string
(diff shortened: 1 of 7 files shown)


src/models/yaml-bundle-descriptor.ts
```diff
@@ -22,7 +22,8 @@ export type YamlWidgetDescriptor = {
 }
 
 export type YamlPluginDescriptor = {
-  descriptorVersion: 'v4'
+  name: string
+  descriptorVersion: 'v5'
```
cecchisandrone: why is this hardcoded here?

raxkaynan: I think it was hardcoded there to avoid having to specify another version currently. I'll change it to `string` instead.

cecchisandrone: Yeah I think it's better to rely on the converter to decide the versions as other descriptor types

src/models/yaml-bundle-descriptor.ts
```diff
@@ -35,7 +36,7 @@ export type YamlPluginDescriptor = {
 }
 
 export type YamlBundleDescriptor = {
-  code: string
+  name: string
   version: string
```
cecchisandrone: descriptorVersion

src/models/yaml-bundle-descriptor.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from './bundle-descriptor'
 
 export type YamlWidgetDescriptor = {
-  code: string
+  name: string
   titles: { [lang: string]: string }
   group: string
   version: string
```
cecchisandrone: descriptorVersion

cecchisandrone (approved):

PR #59, opened 2022-06-16
ENG-3833 amend bundle thumbnail option

jeffgo10:


src/services/bundle-descriptor-converter-service.ts
```diff
@@ -28,17 +28,33 @@ const WIDGET_DESCRIPTOR_VERSION = 'v2'
 const BUNDLE_DESCRIPTOR_VERSION = 'v2'
 const BUNDLE_DESCRIPTOR_NAME = 'descriptor' + DESCRIPTOR_EXTENSION
 
+export enum ThumbnailStatusMessage {
+  NONE = '',
```
cecchisandrone: do we really need this? Can't we use NO_THUMBNAIL?

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -28,17 +28,33 @@ const WIDGET_DESCRIPTOR_VERSION = 'v2'
...
+  FILESIZE_EXCEED = 'Thumbnail filesize is too big. Make sure the thumbnail filesize is 100 KB or below.',
```
cecchisandrone: Bundle thumbnail file ignored because the file size exceeds 100 KB

cecchisandrone: FILESIZE_EXCEEDED

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -28,17 +28,33 @@ const WIDGET_DESCRIPTOR_VERSION = 'v2'
...
+  OK = 'passed',
+  NO_THUMBNAIL = 'No thumbnail found. Please provide a thumbnail in PNG or JPEG format (e.g. `thumbnail.png`) and place it under the root of your bundle directory'
```
cecchisandrone: No bundle thumbnail file found. Please provide a `thumbnail.(png|jpg|jpeg)` and place it under the root of your bundle directory

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -65,6 +81,45 @@ export class BundleDescriptorConverterService {
...
+      this.thumbnail = {
+        ...thumbInfo,
+        status: ThumbnailStatusMessage.NO_THUMBNAIL
+      }
+      return this.thumbnail
+    }
+
+    const stats = fs.statSync(thumbnail)
+    const size = stats.size / 1024
+    const base64 = Buffer.from(fs.readFileSync(thumbnail)).toString('base64')
```
cecchisandrone: this is not enough, you have to provide the URI prefix like: `data:image/png;base64,<base64>`. Note that you have to set `image/jpeg` for both jpg and jpeg extensions

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -65,6 +81,45 @@ export class BundleDescriptorConverterService {
     this.generateBundleYamlDescriptor(bundleDescriptor)
   }
 
+  public processThumbnail(): BundleThumbnailInfo {
```
cecchisandrone: I suggest to extract thumbnail logic in a separate service

src/services/bundle-descriptor-converter-service.ts
jeffgo10: done. changed it to NO_THUMBNAIL

src/services/bundle-descriptor-converter-service.ts
jeffgo10: done

jeffgo10: done

src/services/bundle-descriptor-converter-service.ts
jeffgo10: done

src/services/bundle-descriptor-converter-service.ts
jeffgo10: done

src/services/bundle-descriptor-converter-service.ts
jeffgo10: logic now moved to new service `bundle-thumbnail-service.ts`

src/services/bundle-thumbnail-service.ts
```diff
@@ -0,0 +1,72 @@
...
+  private readonly bundleDirectory: string
+  private size: number
+  private status: ThumbnailStatusMessage
+  private base64: string
+  private path: string
+
+  constructor() {
+    this.bundleDirectory = process.cwd()
+    this.path = ''
+    this.base64 = ''
```
zonia3000: We could set `base64?: string` and `path?: string` both in `BundleThumbnailService` and `BundleThumbnailInfo`, so we can avoid this initialization and return undefined instead of `''` in `findThumbnail()`

test/services/bundle-thumbnail-service.test.ts
```diff
@@ -0,0 +1,86 @@
...
+} from '../../src/services/bundle-thumbnail-service'
+
+describe('bundle thumbnail service', () => {
+  const tempDirHelper = new TempDirHelper(__filename)
+  let bundleDir: string
+  let stubExistsSync: sinon.SinonStub
+
+  before(() => {
+    bundleDir = path.resolve(tempDirHelper.tmpDir, 'test-bundle')
+    fs.mkdirSync(bundleDir, { recursive: true })
```
zonia3000: If you want you can use
```
bundleDir = tempDirHelper.createInitializedBundleDir('test-bundle')
```

src/services/bundle-thumbnail-service.ts
jeffgo10: done

test/services/bundle-thumbnail-service.test.ts
jeffgo10: done

src/services/bundle-thumbnail-service.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor() {
+    this.bundleDirectory = process.cwd()
+    this.size = 0
+    this.status = ThumbnailStatusMessage.NO_THUMBNAIL
+  }
+
+  public processThumbnail(): void {
+    this.path = this.findThumbnail()
+    if (this.path === '') {
```
zonia3000: Check for undefined here

src/services/bundle-thumbnail-service.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  private findThumbnail(): string {
+    for (const ext of ['png', 'jpg', 'jpeg']) {
+      const thumb = path.resolve(this.bundleDirectory, `thumbnail.${ext}`)
+      if (fs.existsSync(thumb)) {
+        return thumb
+      }
+    }
+
+    return ''
```
zonia3000: Returns undefined here

src/services/bundle-thumbnail-service.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public getThumbnailInfo(): BundleThumbnailInfo {
+    const info: BundleThumbnailInfo = {
+      size: this.size,
+      status: this.status
+    }
+    if (this.path) info.path = this.path
+    if (this.base64) info.base64 = this.base64
```
zonia3000, line 68: No need for the if in these cases

jeffgo10: @zonia3000 I wanted not to have `path` and `base64` properties if their values are `undefined`. If I put them like this:
```
obj = { path: this.path, base64: this.base64 }
```
properties `path` and `base64` would still exist in `obj` even if their values are `undefined`

zonia3000: Ah, I see... Thanks for the clarification!

src/services/bundle-thumbnail-service.ts
jeffgo10: done

src/services/bundle-thumbnail-service.ts
jeffgo10: done

cecchisandrone (approved):

zonia3000 (approved):

PR #63, opened 2022-06-17
ENG-3822 Add contextParams to widget micro frontend

raxkaynan:
* Adds `contextParams` property to widget micro frontend
* Improves constraints including error messages
  - The key-matching logic implemented in the union type validation narrows down the errors to the ones that most closely match the expected union type member/s
* Fixes typescript-eslint warning


src/models/bundle-descriptor-constraints.ts
```diff
@@ -209,179 +214,239 @@ const MICROSERVICE_CONSTRAINTS: ObjectConstraints<Microservice> = {
...
+  contextParams: {
+    isArray: true,
+    required: false,
+    type: 'string',
+    validators: [values(WidgetContextParam)]
+  }
+}
+
+const WIDGETCONFIG_MICROFRONTEND_CONSTRAINTS: ObjectConstraints<WidgetConfigMicroFrontend> = {
+  name: {
```
cecchisandrone, line 274: Can't we group all constrainst common to all MFE types (name, type, ..) without repeating them?

raxkaynan: It's quite tricky to do that due to how the `ObjectConstraints` type is currently structured. It can probably be done via something like:
```
type BaseMicroFrontendConstraints = ObjectConstraints<BaseMicroFrontendConstraints> // the type to use for the common properties
type WidgetMicroFrontendConstraints = BaseMicroFrontendConstraints & ObjectConstraints<WidgetMicroFrontend>
...
```
 Any thoughts @zonia3000?

zonia3000: No smart ideas at the moment. I played with this for a while and I agree that it is trickier than it seems at first sight...

cecchisandrone: Ok let's stay with this if it's too much complicated

cecchisandrone (approved):

PR #71, opened 2022-06-22
ENG-3824 configMfe attribute validations for microservices in bundle descriptor

jeffgo10:


src/models/component.ts
```diff
@@ -24,6 +24,7 @@ export type Component<T extends ComponentType> = {
   name: string
   stack: StackFor<T>
   type: T
+  others?: any
```
cecchisandrone: I don't like this, it's too vague. Can't we just map only the props we need explicitly? mfeType, configMfe if I got it well

src/services/component-service.ts
```diff
@@ -133,6 +134,29 @@ export class ComponentService {
...
+        mfe.others?.configMfe &&
+        mfe.others.configMfe.slice(0, 9) !== 'internal:' &&
+        allMfesWithTypes.findIndex(
+          ({ name, type }) =>
+            mfe.others.configMfe === name &&
+            type === MicroFrontendType.WidgetConfig
+        ) === -1
+      ) {
+        throw new CLIError(
+          `Config MFE ${mfe.others.configMfe} must be at least an internal widget or must exist among your micro frontends with type widget-config`
```
cecchisandrone: `configMfe value ${mfe.others.configMfe} for MFE ${mfe} must be an existing widget-config MFE in the project or a platform provided config MFE, defined with the internal:<name> syntax`

src/models/component.ts
jeffgo10: done. I removed `others` props and added props `mfeType` and `configMfe` (limited to ComponentType.MICROFRONTEND only)

src/services/component-service.ts
jeffgo10: done

cecchisandrone (approved):

PR #77, opened 2022-06-24
ENG-3865 Run all microservices

alepintus:


src/commands/run.ts
```diff
@@ -15,12 +20,31 @@ export default class Run extends BaseBuildCommand {
...
   public async run(): Promise<void> {
     BundleService.isValidBundleProject()
-    const { args } = await this.parse(Run)
+    const { args, flags } = await this.parse(Run)
     const componentService = new ComponentService()
 
-    if (args.name === undefined) {
+    if (Object.keys(flags).length > 0 && args.name !== undefined) {
+      this.error(
+        `Run failed, please use one flag or write the component name as argument`,
```
cecchisandrone: Bad arguments. Please use the component name as argument or one of the available flags, not both

src/commands/run.ts
```diff
@@ -15,12 +20,31 @@ export default class Run extends BaseBuildCommand {
...
+        { exit: 1 }
+      )
+    }
+
+    if (flags['all-ms']) {
+      CliUx.ux.action.start(
+        `Running all microservices. Press ctrl + c to exit.`
+      )
+      await this.runAllComponents(Phase.Run, ComponentType.MICROSERVICE)
+    }else {
```
cecchisandrone: bad formatting?

src/commands/run.ts
```diff
@@ -15,12 +20,31 @@ export default class Run extends BaseBuildCommand {
...
+     if (args.name === undefined) {
       this.error(`Run failed, missing required arg name`, { exit: 1 })
```
cecchisandrone: Bad arguments. Please use the component name as argument or one of the available flags

src/commands/run.ts
```diff
@@ -45,6 +69,32 @@ export default class Run extends BaseBuildCommand {
...
 
-    CliUx.ux.action.stop()
+    }
   }
+
+  public async runAllComponents(
+    commandPhase: Phase,
+    componentType?: ComponentType
+  ): Promise<void> {
+    this.log(color.bold.blue('Run components...'))
```
cecchisandrone: Running components `component1`, `component2`

src/utils.ts
```diff
@@ -11,3 +13,35 @@ export class InMemoryWritable extends Writable {
     next()
   }
 }
+
+export class OutputColorWritable extends Writable {
```
cecchisandrone: maybe better `ColorizedWritable`?

alepintus: done

src/commands/run.ts
alepintus: I added the component type , I suggest to address this change (add the list of components) in another task

src/commands/run.ts
alepintus: The user can pass 3 flags as well, so I generalized the message to "Bad arguments. Use the component name as argument or one of the available flags" I also refactored a bit the code to have the same error messages for the build command

src/commands/run.ts
alepintus: Reformatted with prettier

src/commands/run.ts
alepintus: done

cecchisandrone (approved):

PR #84, opened 2022-06-28
ENG-3864 Run all micro frontends

alepintus:


src/commands/run.ts
```diff
@@ -32,7 +39,12 @@ export default class Run extends BaseExecutionCommand {
 
     this.validateInputs(Object.keys(flags).length, args.name)
 
-    if (flags['all-ms']) {
+    if (flags['all-mfe']) {
+      CliUx.ux.action.start(
+        `Running all micro frontends. Press ctrl + c to exit.`
+      )
+      await this.runAllComponents(Phase.Run, ComponentType.MICROFRONTEND)
+    } else if (flags['all-ms']) {
```
raxkaynan, line 49: If both `flags['all-mfe']` and `flags['all-ms']` are set, shouldn't it throw a specific error? If so, the `exclusive` flag property can be used for it.

raxkaynan (changes requested):

src/commands/run.ts
alepintus: @raxkaynan thanks for the suggestion
@cecchisandrone WDYT? using the exclusive flag property the error message would be something like this:
`--all-mfe= cannot also be provided when using --all-ms= `

cecchisandrone: yes seems reasonable

alepintus: done

raxkaynan (approved):

src/commands/run.ts
```diff
@@ -89,7 +103,10 @@ export default class Run extends BaseExecutionCommand {
       component => new ColorizedWritable(component.name, maxPrefixLength)
     )
 
-    const executorService = new ParallelProcessExecutorService(executionOptions)
+    const executorService = new ParallelProcessExecutorService(
+      executionOptions,
+      10
```
zonia3000, line 108: Didn't we say to put this equal to the number of components?

alepintus: @zonia3000 I'll commit that change in the next PR (--all)

PR #97, opened 2022-07-11
ENG-3939 modify creation of default service yaml files to execute on svc enable command instead during init phase

jeffgo10:


src/services/create-default-svc-service.ts
```diff
@@ -0,0 +1,150 @@
+import * as fs from 'node:fs'
```
raxkaynan: Not sure if a separate service class is needed here. Couldn't this be just merged into `svc-service`?

cecchisandrone: I may agree in splitting them as they have different responsibilities and class could be difficult to read

cecchisandrone (approved):

PR #99, opened 2022-07-13
ENG-3931 Fixed Windows paths issue

zonia3000:


raxkaynan (approved):

src/services/docker-service.ts
```diff
@@ -196,25 +196,40 @@ export class DockerService {
     return images
   }
 
-  public static async login(
-    registry: string = DEFAULT_DOCKER_REGISTRY
-  ): Promise<void> {
+  public static async tryLogin(
```
cecchisandrone: I would name it `checkAuthentication`

cecchisandrone (approved):

PR #105, opened 2022-07-20
ENG-3979 add confirmation when removing references of widget config

jeffgo10:


src/commands/mfe/rm.ts
```diff
@@ -22,6 +25,48 @@ export default class Rm extends Command {
...
       new MicroFrontendService()
+    const mfeReferences = microFrontendService.findWidgetConfigReferences(
+      args.name
+    )
+
+    if (mfeReferences.length > 0) {
+      const { removeConfigs } = await inquirer.prompt([
+        {
+          name: 'removeConfigs',
+          message: `Reference to ${args.name} will be removed from this bundle's microfrontend widgets. Do you wish to continue?`,
```
zonia3000: Could you please replace "this bundle's microfrontend widgets" with the actual list of widgets names? We agreed that when we discussed the AC.

src/services/microfrontend-service.ts
```diff
@@ -90,6 +90,33 @@ export class MicroFrontendService {
     return microfrontends.find(({ name }) => name === mfeName)
   }
 
+  private isConfigMfeSiblingWidget(mfe: MicroFrontend): boolean {
+    return (
+      'configMfe' in mfe &&
+      mfe.configMfe !== undefined &&
+      mfe.configMfe.slice(0, 9) !== 'internal:'
```
zonia3000: Maybe we can extract `internal:` to a constant, shared also with `component-service.ts`.

src/commands/mfe/rm.ts
jeffgo10: done

src/services/microfrontend-service.ts
jeffgo10: done

zonia3000 (approved):

PR #107, opened 2022-07-29
ENG-4023 change prefix validator in configMfe for platform widgets from internal to global

jeffgo10:


zonia3000 (changes requested):
Hi @jeffgo10, this is not working because the validation check uses the following logic, but the string length is changed:
```
mfe.configMfe.slice(0, 9) !== PREFIX_INTERNAL_WIDGET
```
I suggest to change it to:
```
!mfe.configMfe.startsWith(PREFIX_INTERNAL_WIDGET)
```
Moreover we should call the variable `PREFIX_GLOBAL_WIDGET`
It also looks that `internal:` is still hardcoded in the error message at `checkConfigMfes()` inside `component-service.ts`
Please add a unit test for this case.

jeffgo10:
> Hi @jeffgo10, this is not working because the validation check uses the following logic, but the string length is changed:
> 
> ```
> mfe.configMfe.slice(0, 9) !== PREFIX_INTERNAL_WIDGET
> ```
> 
> I suggest to change it to:
> 
> ```
> !mfe.configMfe.startsWith(PREFIX_INTERNAL_WIDGET)
> ```
> 
> Moreover we should call the variable `PREFIX_GLOBAL_WIDGET` It also looks that `internal:` is still hardcoded in the error message at `checkConfigMfes()` inside `component-service.ts` Please add a unit test for this case.

@zonia3000 done. pls re-review

zonia3000 (approved):

PR #108, opened 2022-07-29
ENG-4022 Add missing micro frontend fields

raxkaynan:
* Adds parentName, parentCode, paramsDefaults, and customUi micro frontend fields
* Adds `exclusive()` validation function used for parentCode and parentName validation
* Adds params field to widget config micro frontend
* Implements logic that handles how customUiPath is generated

### DIFF
--- a/src/models/bundle-descriptor-constraints.ts
+++ b/src/models/bundle-descriptor-constraints.ts
@@ -326,10 +326,6 @@ const WIDGET_MICROFRONTEND_CONSTRAINTS: ObjectConstraints<WidgetMicroFrontend> =
       required: false,
       children: PARAM_CONSTRAINTS
     },
-    customUi: {
-      required: false,
-      type: 'string'
-    },
     parentName: {
       required: false,
       type: 'string'
@@ -396,10 +392,6 @@ const WIDGETCONFIG_MICROFRONTEND_CONSTRAINTS: ObjectConstraints<WidgetConfigMicr
       isArray: true,
       required: false,
       children: PARAM_CONSTRAINTS
-    },
-    customUi: {
-      required: false,
-      type: 'string'
     },
     parentName: {
       required: false,
@@ -475,10 +467,6 @@ const APPBUILDER_MICROFRONTEND_CONSTRAINTS: Array<
       required: false,
       children: PARAM_CONSTRAINTS
     },
-    customUi: {
-      required: false,
-      type: 'string'
-    },
     parentName: {
       required: false,
       type: 'string'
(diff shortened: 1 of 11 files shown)


cecchisandrone:
image
Please fix this

src/models/bundle-descriptor-constraints.ts
```diff
@@ -324,6 +325,23 @@ const WIDGET_MICROFRONTEND_CONSTRAINTS: ObjectConstraints<WidgetMicroFrontend> =
...
+    },
+    parentName: {
+      required: false,
+      type: 'string'
+    },
+    parentCode: {
+      required: false,
+      type: 'string'
+    },
+    paramsDefaults: {
```
cecchisandrone, line 337: can't we put default inside already existing `params` object? it seems cleaner to me

src/models/bundle-descriptor-constraints.ts
```diff
@@ -324,6 +325,23 @@ const WIDGET_MICROFRONTEND_CONSTRAINTS: ObjectConstraints<WidgetMicroFrontend> =
       isArray: true,
       required: false,
       children: PARAM_CONSTRAINTS
+    },
+    customUi: {
```
cecchisandrone: I don't see a big benefit in supporting customUi snippets here, we may have problems with readability. I just agreed with @w-caffiero-entando and @eugeniosant to remove it

src/paths.ts
```diff
@@ -16,3 +16,4 @@ export const BUILD_FOLDER = 'build'
 export const DESCRIPTOR_EXTENSION = '.yaml'
 export const BUNDLE_DESCRIPTOR_NAME = 'descriptor' + DESCRIPTOR_EXTENSION
 export const GITKEEP_FILE = '.gitkeep'
+export const CUSTOM_UI_EXTENSION = '.ftl'
```
cecchisandrone: CUSTOM_WIDGET_TEMPLATE_EXTENSION

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -85,11 +87,31 @@ export class BundleDescriptorConverterService {
...
       | YamlWidgetDescriptor
       | YamlWidgetConfigDescriptor
-      | YamlAppBuilderWidgetDescriptor =
-      this.mapMicroFrontendToYamlDescriptor(microFrontend)
+      | YamlAppBuilderWidgetDescriptor = this.mapMicroFrontendToYamlDescriptor(
+      microFrontend,
+      customUiFileExists
+    )
+
+    if (customUiFileExists) {
```
cecchisandrone, line 105: In case FTL file is found, I would add an info log printing: "Custom widget template FTL found, including it"

src/models/bundle-descriptor-constraints.ts
raxkaynan: but in the yaml descriptor, should it still be separate?

cecchisandrone: I just realized that `params` and `paramsDefault` are not exactly the same list. The second can contain "hidden" params so let's keep it as it is

cecchisandrone (approved):

PR #116, opened 2022-08-11
ENG-4087 add warning message on cloning bundle without descriptor json

jeffgo10:


src/services/initializer-service.ts
```diff
@@ -112,15 +115,19 @@ export class InitializerService {
...
+        this.filesys.getBundleDirectory()
+      )
+      const descriptor = bundleDescriptorService.getBundleDescriptor()
+      bundleDescriptorService.writeBundleDescriptor({
+        ...descriptor,
+        name,
+        version
+      })
+    } catch {
+      throw new CLIError(ERROR_NO_BUNDLE_DESCRIPTOR)
```
zonia3000: It is also possible to use `Errors.warn(color.bold.yellow(ERROR_NO_BUNDLE_DESCRIPTOR))` (`Errors` from `'@oclif/core'`), that can be used in services instead of the `this.warn` available in commands classes.
In this way no exception is thrown, avoiding the try/catch+if in the command, and the message is simply print on the stderr.

jeffgo10: done

src/services/initializer-service.ts
```diff
@@ -112,15 +114,25 @@ export class InitializerService {
...
+        this.filesys.getBundleDirectory()
+      )
+      const descriptor = bundleDescriptorService.getBundleDescriptor()
+      bundleDescriptorService.writeBundleDescriptor({
+        ...descriptor,
+        name,
+        version
+      })
+    } catch {
+      if (process.env.ENTANDO_BUNDLE_CLI_NO_JSON_WARNING === 'false') {
```
cecchisandrone: it looks a bit too generic catch. let's explicitly check for entando.json existence

jeffgo10: done. I added logic on checking the error.message to ensure that it is catching with the right error

cecchisandrone (approved):

PR #124, opened 2022-09-12
ENG-4165 Simplify secret reference declaration in entando.json

alepintus:


src/models/bundle-descriptor.ts
```diff
@@ -1,19 +1,35 @@
...
+
+export type SecretEnvironmentVariable = {
+  name: string
+  secretKeyRef: {
+    name: string
+    key: string
+  }
+}
+
+export type SecretEnvironmentVariableYaml = {
```
cecchisandrone: This should not be here but in yaml-bundle-descriptor, no?

cecchisandrone (approved):

src/models/bundle-descriptor.ts
alepintus: done

PR #125, opened 2022-09-16
ENG-4196 Add category field in MFE of  type widget

alepintus:


src/models/yaml-bundle-descriptor.ts
```diff
@@ -54,6 +54,7 @@ export type YamlWidgetDescriptor =
     configMfe?: string
     contextParams?: string[]
     nav?: Nav[]
+    category?: string
```
cecchisandrone: this is better to be widgetCategory as it is the same attribute we have engine side. in this case ecr will just use the same attribute

alepintus: done

cecchisandrone (approved):

PR #129, opened 2022-11-24
ENG-4176: added precise error location info when validating descriptor

antromeo:


zonia3000 (changes requested):
If the error is at the last element of the line (i.e. missing comma at the end of one field) it highlights the line after the error:
image

Line numbers usually start from 1:
image

I would also add some padding to the other 2 lines to align the output.

PR #132, opened 2022-12-02
ENG-4356: added support compose like a command of docker

antromeo:


src/services/svc-service.ts
```diff
@@ -195,12 +193,18 @@ export class SvcService {
       workDir: this.parentDirectory
     }
 
-    // detect the presence docker-compose-plugin
+    // detects the presence docker-compose-plugin
+    SvcService.debug(`detecting the presence of docker-compose-plugin`)
     const code = await ProcessExecutorService.executeProcess(options)
+    let foundcomposeplugin
```
cecchisandrone: composePluginFound

cecchisandrone (approved):

PR #133, opened 2022-12-05
ENG-4381 Process exit with code 1 in case of build errors

alepintus:


antromeo (approved):

src/commands/base-build.ts
```diff
@@ -124,7 +124,7 @@ export abstract class BaseBuildCommand extends BaseExecutionCommand {
       }
 
       errorMessage += 'See log files for more information'
-      this.error(errorMessage)
+      this.error(errorMessage, { exit: 1 })
```
cecchisandrone, line 127: what about pack? I know the ticket was mentioning build only but we have to keep consistent behavior also there

alepintus: I did some tests yesterday and the error is properly propagated for the pack command as well

cecchisandrone (approved):

PR #134, opened 2022-12-06
ENG-4394 Added --stdout flag to build and pack commands

zonia3000:


src/commands/build.ts
```diff
@@ -33,23 +33,34 @@ export default class Build extends BaseBuildCommand {
     all: Flags.boolean({
       description: 'Build all the bundle components',
       exclusive: ['all-ms', 'all-mfe']
+    }),
+    stdout: Flags.boolean({
+      description: 'Log build output to standard output'
```
cecchisandrone: Print build output to stout instead of files

cecchisandrone (approved):

PR #150, opened 2023-03-16
Multi-Arch BUILDX CLI tidbits

lward27:
Here were the "touch points and learnings" that we came across while shooting for building a multi-arch container using the CLI. I don't expect this pull request to be merged, but wanted to share our findings.  Please note, our private registry is hard coded, - again, this PR is to just serve as a reference and act as a way to share our notes.

- BuildX requires some additional configuration regarding setting the buildx backend
  - looks like this: `docker buildx create --use --driver-opt network=host --buildkitd-flags '--debug --allow-insecure-entitlement network.host'`
- Entando work flow is like: pack and then publish, which didn't fit great with the buildx model.
  - without a registry to put the multiarch images in, one image would be stored locally and the other would be cached... To work around this, we made the pack command push to our registry with --push, which goes slightly outside the Entando work flow.
- We quickly realized that any remaining distance down this path would likely include the custom operator and application code changes, so we decided to stop there.

### DIFF
--- a/src/services/docker-service.ts
+++ b/src/services/docker-service.ts
@@ -81,7 +81,7 @@ export class DockerService {
       options.name,
       options.tag
     )
-    return `${DOCKER_COMMAND} buildx build --push --platform "linux/arm64,linux/amd64" -f ${dockerfile} -t ${dockerImageName} .`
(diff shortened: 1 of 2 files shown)


pietrangelo:
Thanks for your collaboration @lward27 we really appreciate it. We are working to add support for multi-architectures in our platform and as already you noted, we have a lot to cover. 
Thanks again!!

PR #151, opened 2023-03-28
ENG-4715 generate-cr can select prod and dev tags

firegloves:


src/commands/generate-cr.ts
```diff
@@ -130,13 +151,15 @@ export default class GenerateCr extends Command {
       this.error(`No tags found for the Docker image ${image}`)
     }
 
+    const filteredTags = this.filterTagsByAllowedTypes(flags.tagtypes || ['dev', 'prod'], tags);
```
antromeo: could the OR be replaced by nullish coalescing operator?

firegloves: @antromeo yes, done. Thanks

antromeo (approved):

test/commands/generate-cr.test.ts
```diff
@@ -335,4 +330,56 @@ describe('generate-cr', () => {
...
+function createStubDigestsExecutor() {
+  return new (class extends StubParallelProcessExecutorService {
+    async getDigests(): Promise<Map<string, string>> {
+      await super.execute()
+      const tagsWithDigests = new Map()
+      tagsWithDigests.set('0.0.2', 'sha256:abcd')
+      tagsWithDigests.set('0.0.1', 'sha256:1234')
+      return tagsWithDigests
+    }
+  })([0, 0])
```
zonia3000: To avoid certain random issues we experienced in the past with events in tests the number of stubbed results passed to the constructor should be the same of the expected progress bar steps.
I suggest passing a parameter like `resultsCount: number` to the `createStubDigestsExecutor()` function and use something like `new Array(resultsCount).fill(0)` instead of `([0,0])` here.

This is also the reason why in many tests we also added checks like `expect(ctx.stderr).contain('2/2')`, to check that the progress bar completes as expected.

zonia3000 (changes requested):

PR #165, opened 2023-05-24
ENG-4833 - fix MySQL template yaml

MEM2677:
This simple fix should fix the login to MySQL problem on local development environment.


firegloves (approved):
LGTM

PR #174, opened 2023-06-20
ENG-4951: conversion of service files

antromeo:


antromeo:
Wait ENG-4950

avdev4j:
Is the svc files supposed to be in the bundle? Because there only present in the src project that not always contains the bundle descriptors. So, I was thinking we focus on the delivered v1 bundle that doesn't contain the svc, for sure

antromeo:
> Is the svc files supposed to be in the bundle? Because there only present in the src project that not always contains the bundle descriptors. So, I was thinking we focus on the delivered v1 bundle that doesn't contain the svc, for sure

In Bundle URI, you probably won't find them, but that's an optional step where the svc path folder is a user input and the user can specify the path if they are used locally wherever they are.
In next steps, if we choose to support Source URL, this optional step will become more useful than now.

CC: @firegloves

avdev4j:
@antromeo thanks for the further information. It makes sense now, I would suggest making the prompt more obvious (maybe give more details in the prompt sentence). As you said, if we also support the sources Bundle, it will be more useful because we are almost sure to have them.

Another solution would be to ask the user if he/she wants to add the svc and then ask for the path.

antromeo:
@avdev4j in my opinion two questions are maybe too verbose.
Yes the sentence could be improved in something like "Enter the svc path if you are using svc files".
Let me know for your suggestions

src/commands/convert.ts
```diff
@@ -63,9 +70,22 @@ export default class Convert extends Command {
       register.push(...msStatus)
     }
 
+    // adds services
+    servicePath = servicePath ?? await CliUx.ux.prompt(
+      'Services path (optional)',
```
avdev4j:
```suggestion
      'Please provide the path to the service files source code (optional)',
```

avdev4j (changes requested):

src/commands/convert.ts
avdev4j: suggestion from @jyunmitch

antromeo: @avdev4j, @jyunmitch  
just a note: what do you think if we took away the words "source code"? because docker compose files are actually yaml files (no source)

avdev4j (approved):

alepintus (approved):

src/commands/convert.ts
jyunmitch: Does this work? "Please provide the path to the YAML service files (optional)"
or "Please provide the path to the Docker Compose service files (optional)"

PR #179, opened 2023-06-28
ENG-4991 add health check ingress

avdev4j:


src/models/bundle-descriptor-constraints.ts
```diff
@@ -242,6 +242,10 @@ export const MICROSERVICE_CONSTRAINTS: ObjectConstraints<Microservice> = {
     required: true,
     type: 'string'
   },
+  healthCheckIngress: {
+    required: true,
```
antromeo: Should the field healthCheckIngress be required? I think no for backwards compatibility.
In this case, in old bundles, any commands you execute returns the following error. 
```
Error: entando.json is not valid.
Field "healthCheckIngress" is required
```

Furthermore, if you want to bind the values to the two enumeratives, you should add:
```
validators: [values(HealthCheckIngress)]
```

src/services/microservice-service.ts
```diff
@@ -55,7 +55,8 @@ export class MicroserviceService {
 
     this.addMicroserviceDescriptor({
       ...ms,
-      healthCheckPath: ms.healthCheckPath ?? DEFAULT_HEALTH_CHECK_PATH
+      healthCheckPath: ms.healthCheckPath ?? DEFAULT_HEALTH_CHECK_PATH,
+      healthCheckIngress: ms.healthCheckIngress ?? HealthCheckIngress.Canonical
```
antromeo, line 58: ok for input but in output when you run `ent bundle pack` you should find your property in output (so`.entando/output/plugins/plugin-desc.yaml`) in that way the CM can processing the CR when it is applied.
try and let me know

src/models/bundle-descriptor-constraints.ts
avdev4j: @antromeo I checked the ticket and it is not defined if the property should be required or not. So, we can update that!

src/services/microservice-service.ts
avdev4j: let's discuss that later. Not sure to get it

avdev4j: done

src/models/bundle-descriptor-constraints.ts
avdev4j: done

antromeo (approved):

PR #182, opened 2023-07-12
ENG-5039 backport eng 4883

avdev4j:


antromeo (approved):
lgtm, only later I think we should make the fix described in ENG-5012 in both versions
@avdev4j @firegloves

PR #183, opened 2023-07-13
ENG-4496 Update Enable debug mode messages handling

alepintus:


src/services/docker-service.ts
```diff
@@ -526,27 +566,32 @@ export class DockerService {
           }
 
           DockerService.debug(output)
+
+          const errorMsg= debugEnabled ? errUnableToExtractDigestMsg : `${errUnableToExtractDigestMsg}${enableDebugRetrievedContentMsg}`
```
firegloves, line 570: should we add a white space between `${errUnableToExtractDigestMsg}${enableDebugRetrievedContentMsg}`?

src/utils.ts
```diff
@@ -8,6 +8,11 @@ export const EXIT_CODES = {
   SUCCESS: 0,
   GENERIC_ERROR: 1
 }
+export function isDebugEnabled(): boolean {
+  return (
+    process.env.ENTANDO_CLI_DEBUG === 'true' ?? false
```
firegloves, line 13: is the `?? false` required?

firegloves (changes requested):

src/utils.ts
alepintus: Yes, we talked about that with @antromeo , in this situation it's better to add a fallback value using the Nullish coalescing "??" Operator

avdev4j: I think so, ENTANDO_CLI_DEBUG is a string, and that given code does "if ENTANDO_CLI_DEBUG value is "true" return true or if it is something else or null return false".

It could be 
`return process.env.ENTANDO_CLI_DEBUG === "true" ? true : false` 
or 
```
if(process.env.ENTANDO_CLI_DEBUG === "true") {
    return true 
} else {
    return false
}
```

The current code seems to be more concised.

src/services/docker-service.ts
alepintus: IMHO The space between the two strings variables it's not necessary in this case because the first string have a "\n" at the end.

src/utils.ts
antromeo: as @avdev4j said , `return process.env.ENTANDO_CLI_DEBUG === "true" ? true : false` it looks nicer to me

firegloves (approved):
LGTM

PR #191, opened 2024-01-15
ENG-5365 Add bundleDescriptorVersion field to improve backwards compatibility

alepintus:


src/services/bundle-descriptor-converter-service.ts
```diff
@@ -91,7 +93,7 @@ export class BundleDescriptorConverterService {
     )
   }
 
-  private generateMicroFrontendYamlDescriptor(microFrontend: MicroFrontend) {
+  private generateMicroFrontendYamlDescriptor(microFrontend: MicroFrontend, bundleDescriptorVersion: BundleDescriptorVersion | undefined) {
```
firegloves, line 96: do we want to use the `!` operator?

src/services/bundle-descriptor-converter-service.ts
```diff
@@ -145,7 +148,8 @@ export class BundleDescriptorConverterService {
 
   private mapMicroFrontendToYamlDescriptor(
     microFrontend: MicroFrontend,
-    customUiFileExists: boolean
+    customUiFileExists: boolean,
+    bundleDescriptorVersion: BundleDescriptorVersion | undefined
```
firegloves, line 152: do we want to use the `!` operator here too?

firegloves (approved):
LGTM

PR #193, opened 2024-01-29
ENG-5448 Add overwriteTenants flag to generate-cr

alepintus:


src/commands/generate-cr.ts
```diff
@@ -74,6 +74,16 @@ export default class GenerateCr extends Command {
       multiple: true,
       description: 'Accepted tag types, comma separated values. Accepted values are ' + [...GenerateCr.tagTypeStrategies.keys()].join(", ")
     }),
+    overwriteTenants: Flags.boolean({
```
antromeo, line 77: could it just be called `overwrite`?

src/commands/generate-cr.ts
```diff
@@ -74,6 +74,16 @@ export default class GenerateCr extends Command {
...
+      char: 'v',
+      description: 'Overwrite the bundle cr tenants list with the value passed in the tenants parameter',
+      dependsOn: ['tenants']
+    }),
+    forceOverwriteTenants: Flags.boolean({
```
antromeo, line 81: you could remove dependsOn in `force` and reuse it

src/commands/generate-cr.ts
```diff
@@ -201,11 +228,21 @@ export default class GenerateCr extends Command {
...
+    if(!flags.overwriteTenants) {
+      deployedBundleTenants = await MultiTenantsService.getEntandoDeBundleTenants(yamlDescriptor.name, image);
+    }
 
-    const deployedBundleTenants = await MultiTenantsService.getEntandoDeBundleTenants(yamlDescriptor.name,image);
-
-    // eslint-disable-next-line new-cap
-    tenants = [...new Set([...deployedBundleTenants,...filteredCmdTenants])].sort(Intl.Collator().compare)
+    // Force the tenants metadata using the list passed in tenants parameter skipping the tenant-secrets check
+    if (flags.overwriteTenants) {
```
antromeo: you can put them in `and`

src/commands/generate-cr.ts
```diff
@@ -74,6 +74,16 @@ export default class GenerateCr extends Command {
       multiple: true,
       description: 'Accepted tag types, comma separated values. Accepted values are ' + [...GenerateCr.tagTypeStrategies.keys()].join(", ")
     }),
+    overwriteTenants: Flags.boolean({
+      char: 'v',
```
antromeo: we could remove the chars but that's just my opinion

antromeo (commented):
Note that the examples contain some errors such as missing commas and missing dashes

src/commands/generate-cr.ts
alepintus: Hi @antromeo , 
Thanks for the review, I added the word **Tenants** to explain better that this option is only related to multi-tenants environments.

src/commands/generate-cr.ts
alepintus: done

src/commands/generate-cr.ts
alepintus: I added the "char" shortcuts just because all the other flags have it

src/commands/generate-cr.ts
alepintus: The force flag and forceOverwriteTenants do different things, It's better IMHO to have 2 different flags

src/commands/generate-cr.ts
antromeo: since it's not mandatory we can remove them if it isn't the first char of the word in order to avoid putting random chars. Try to see `stdout` flag in pack command.

src/commands/generate-cr.ts
antromeo: in pipeline you don't have standard input, so a force flag to disable the requests to std input is sufficient.
Here, it is like we do:

```
entando-bundle-cli generate-cr --A --disableStdInputForA --B --disableStdInputForB
```
instead of
```
entando-bundle-cli generate-cr --A --B --disableStdInput
```

the first case will introduce an additional flag for each flag that needs to disable std input but we needs just one flag for the command to disable the std input (force).

src/commands/generate-cr.ts
```diff
@@ -40,9 +40,9 @@ export default class GenerateCr extends Command {
     '<%= config.bin %> <%= command.id %> --image=my-org/my-bundle --digest',
     '<%= config.bin %> <%= command.id %> -o my-cr.yml',
     '<%= config.bin %> <%= command.id %> -t prod,dev',
-    '<%= config.bin %> <%= command.id %> -tenants primary',
+    '<%= config.bin %> <%= command.id %> --tenants primary',
     '<%= config.bin %> <%= command.id %> -e primary',
-    '<%= config.bin %> <%= command.id %> -tenants primary tenant1 tenant2',
+    '<%= config.bin %> <%= command.id %> --tenants primary tenant1 tenant2',
```
antromeo: tenants should be comma separated

src/commands/generate-cr.ts
alepintus: Yes, you're right. I'm going to do this change

alepintus: Done

src/commands/generate-cr.ts
alepintus: Done

src/commands/generate-cr.ts
alepintus: Example updated

antromeo (approved):
lgtm

PR #194, opened 2024-02-12
ENG-5457 Add external API claim bundle ref

alepintus:


src/services/api-claim-service.ts
```diff
@@ -43,16 +43,22 @@ export class ApiClaimService {
...
-    const url = `${process.env.ENTANDO_CLI_BASE_URL}${ingressPath}`
+    let bundleId = ''
 
-    this.addApiClaim(mfeName, apiClaim)
-    this.updateMfeConfigApiClaim(mfeName, apiClaim.name, url)
+    if (apiClaim.bundle) {
+      bundleId = BundleService.generateBundleId(apiClaim.bundle)
+
+      const cmService = new CmService()
+      const { ingressPath } = await cmService.getBundleMicroservice(
```
antromeo:
```suggestion
      const { ingressPath } = await new CmService().getBundleMicroservice(
```

src/services/api-claim-service.ts
```diff
@@ -43,16 +43,22 @@ export class ApiClaimService {
...
     }
 
-    const bundleId = BundleService.generateBundleId(apiClaim.bundle)
-    const cmService = new CmService()
-    const { ingressPath } = await cmService.getBundleMicroservice(
-      bundleId,
-      apiClaim.serviceName
-    )
-    const url = `${process.env.ENTANDO_CLI_BASE_URL}${ingressPath}`
+    let bundleId = ''
```
antromeo: we can remove it

src/services/api-claim-service.ts
```diff
@@ -43,16 +43,22 @@ export class ApiClaimService {
...
 
-    this.addApiClaim(mfeName, apiClaim)
-    this.updateMfeConfigApiClaim(mfeName, apiClaim.name, url)
+    if (apiClaim.bundle) {
+      bundleId = BundleService.generateBundleId(apiClaim.bundle)
```
antromeo:
```suggestion
      const bundleId = BundleService.generateBundleId(apiClaim.bundle)
```

src/services/constraints-validator-service.ts
```diff
@@ -216,6 +216,34 @@ export function mutualDependency(
   }
 }
 
+export function mutualExclusive(
```
antromeo, line 219: here we are combining mutual dependency and exclusive, it would be better to have a single responsibility per function. possibly generalized and not just for two fields

src/models/bundle-descriptor-constraints.ts
```diff
@@ -149,16 +150,26 @@ export const API_CLAIMS_CONSTRAINTS: UnionTypeConstraints<
...
+        required: false,
+        type: 'string'
       }
     }
   ],
   validators: [
-    mutualDependency(
-      { key: 'type', value: ApiType.External },
-      { key: 'bundle' }
+    mutualExclusive(
```
antromeo, line 164: it would be nice to have something like

```
 exclusive('bundle', 'bundleReference'),
 fieldDependsOnFields(
      { key: 'type', value: "external" },
      { key: 'bundleReference' },
      { key: 'bundle'})
```

src/services/constraints-validator-service.ts
alepintus: Hi  @antromeo , thanks for the review, in this case this a new kind of validation different from exclusive as well as mutual Dependency, we need to have only one of those fields present in the json but one of these must be present.
This validation was not present in the code base before this PR.

src/services/api-claim-service.ts
alepintus: Done

src/services/api-claim-service.ts
alepintus: Done

src/services/api-claim-service.ts
alepintus: Done

src/services/constraints-validator-service.ts
antromeo: what do you think about 
```
export function fieldDependsOnFields(
  field: Field,
  dependsOnFields: Field[]
): UnionTypeValidator {
  return function (object: any, jsonPath: JsonPath) {
    if (
      (field.value !== undefined && object[field.key] === field.value) ||
      (field.value === undefined && object[field.key] !== undefined)
    ) {
      let message = `Field "${field.key}"`
      const undefinedFields = []
      if (field.value) message += ` with value "${field.value}"`

      for (const dependsOnField of dependsOnFields) {
        if (object[dependsOnField.key] === undefined) {
          undefinedFields.push(dependsOnField.key)
        }
      }

      if (undefinedFields.length === dependsOnFields.length) {
        message += ` requires one of fields ${JSON.stringify(undefinedFields)} to have a value`
        throw new PrioritizedValidationError(message, jsonPath)
      }

    }
  }
}
```
```
export function mutualDependencyOneToMany(
  field: Field,
  dependentFields: Field[]
): UnionTypeValidator {
  return function (...args) {
    fieldDependsOnFields(field, dependentFields)(...args)
    for (const dependentField of dependentFields) {
      fieldDependsOn(dependentField, field)(...args)
    }
  };
}
```
and in validators we can call

```
  exclusive('bundle','bundleReference'),
  mutualDependencyOneToMany( 
    { key: 'type', value: ApiType.External },
    [{ key: 'bundle' },  { key: 'bundleReference' }]
  )
```
in this way we separate the responsabilities of dependency (a field with at least one among many) and exclusivity in two functions and we reuse the already existing `exclusive` function.

antromeo (approved):

